PR #155, opened 2023-01-10
[CHK-706] presa in carico

ciuffagianluca:
When the result from nodo for a payment is a timeout and the payment is taken in charge, the outcome is ok, but the payment result is not still available, for a better user experience the result page should show a coherent message for this situation

#### List of Changes

Added new type of outcome
Evaluation of id status to check this OK peculiar result
Update translation

#### Types of changes

#### Checklist:


src/utils/transactions/TransactionResultUtil.ts
```diff
@@ -19,6 +20,7 @@ export enum OutcomeEnum {
   INVALID_METHOD = "12",
   KO_RETRIABLE = "13",
   INVALID_SESSION = "14",
+  TAKE_IN_CHARGE = "15",
```
giovanniberti, line 23: [bikeshedding]
What about
```suggestion
  TAKEN_IN_CHARGE = "15",
```
?

src/utils/transactions/TransactionResultUtil.ts
```diff
@@ -30,6 +32,7 @@ export enum ViewOutcomeEnum {
   INVALID_CARD = "7",
   CANCELED_BY_USER = "8",
   EXCESSIVE_AMOUNT = "10",
+  TAKE_IN_CHARGE = "15",
```
giovanniberti, line 35: (see above)

giovanniberti (approved):

giovanniberti:
🚀

src/utils/transactions/TransactionResultUtil.ts
ciuffagianluca: That's reasonable

src/utils/transactions/TransactionResultUtil.ts
ciuffagianluca: That's reasonable

src/utils/transactions/TransactionResultUtil.ts
```diff
@@ -243,5 +247,9 @@ export const getOutcomeFromAuthcodeAndIsDirectAcquirer = (
             )
           )
         )
-    )
+    ),
+    (outcomeEnum: OutcomeEnum) =>
+      outcomeEnum === OutcomeEnum.SUCCESS && idStatus === 14
```
infantesimone, line 252: are we sure that the state to consider is 14?

infantesimone (changes requested):
this PR have to go into prod before ecommerce?

PR #165, opened 2023-01-23
fix(pipelines): fix e2e tests path into tests repository

giovanniberti:
#### List of Changes

* fix e2e tests path into tests repository

#### Types of changes

#### Checklist:


Fernando-Granato (commented):
being the main the production photography the tests will fail as the button components are recovered through the id that in this version is not managed in the component FormButtons 😞  
Can we try to bring those changes to the main and run a test what do you think?

infantesimone:
> being the main the production photography the tests will fail as the button components are recovered through the id that in this version is not managed in the component FormButtons 😞 Can we try to bring those changes to the main and run a test what do you think?

@Fernando-Granato can you report the references of the PR with these changes?
cc @giovanniberti

PR #167, opened 2023-01-25
CHK-965-activation-phase

Fernando-Granato:
#### List of Changes

- Modified the call to perform PM to ecommerce payment activation.
- Removed the logic to perform polling for trigger activation state.
- Modified call to perform get transaction by id from PM to ecommerce.

#### Motivation and Context
The old checkout flow has been modified to activate a payment using the new primitives made available by the ecommerce project.

The tests were performed locally 

#### Types of changes

#### Checklist:


package.json
```diff
@@ -7,7 +7,7 @@
...
-    "generate:payment-ecommerce": "rimraf generated/definitions/payment-ecommerce && shx mkdir -p generated/definitions/payment-ecommerce && gen-api-models --api-spec https://raw.githubusercontent.com/pagopa/pagopa-infra/main/src/core/api/checkout/checkout_ecommerce/v1/_openapi.json.tpl --no-strict --out-dir ./generated/definitions/payment-ecommerce --request-types --response-decoders --client",
+    "generate:payment-ecommerce": "rimraf generated/definitions/payment-ecommerce && shx mkdir -p generated/definitions/payment-ecommerce && gen-api-models --api-spec https://raw.githubusercontent.com/pagopa/pagopa-infra/fix-ecommerce-openapi/src/core/api/checkout/checkout_ecommerce/v1/_openapi.json.tpl --no-strict --out-dir ./generated/definitions/payment-ecommerce --request-types --response-decoders --client",
```
Fernando-Granato, line 10: Awaiting the pagopa-infra merge

Infectedmachine (approved):
lgt

giovanniberti (approved):

src/routes/PaymentNoticePage.tsx
```diff
@@ -47,7 +47,7 @@ export default function PaymentNoticePage() {
 
   const onSubmit = React.useCallback(
     async (notice: PaymentFormFields) => {
-      const rptId: RptId = `${notice.cf}${notice.billCode}`;
+      const rptId: RptId = `${notice.cf}${notice.billCode}` as RptId;
```
giovanniberti, line 50: General comment: it would be better to use proper conversions via `io-ts` instead of type assertions (in this case `RptId.decode`

giovanniberti (commented):
m

src/routes/PaymentNoticePage.tsx
infantesimone: something like that, @giovanniberti right?
```
      const rptId: RptId = pipe(
        RptId.decode(JSON.parse(`${notice.cf}${notice.billCode}`)),
        E.fold((_) => onError(ErrorsType.GENERIC_ERROR), identify)
      );
```

infantesimone (approved):
great! great PR to full integration with ecommerce

src/routes/PaymentNoticePage.tsx
giovanniberti: In general, yes! This specific snippet would not typecheck as `onError` returns `void` 🙂, but the general idea is there

Infectedmachine: > In general, yes! This specific snippet would not typecheck as `onError` returns `void` 🙂, but the general idea is there

I dunno about this. if you gonna typecheck on runtime the rptid then you should typecheck every other form in the app. On the other hand the form is already checking for correct string formats, so it's impossible to have errors, the E.fold will be always unreachable. If for some reason it's executed it goes against the provided design which expects the FORM to provide the user with useful information about acceptable string formats, and avoid wrong input. 
So the only point of the type assertion here was to be sure that whenever the code is edited in the future, all the types can overlap and not to check if they do on runtime.

PR #168, opened 2023-01-25
[Chk-967] use authorization request from ecommerce

ciuffagianluca:
The scope of this PR is to change the final step of the payment flow, including the ecommerce api invocation for the payment authorization process

#### List of Changes

- Added property for payment method type code tracking in session storage
- Added ecommerce api `auth-request` invocation

#### Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:


giovanniberti:
Can you add a better title and description? This is a big feature for Checkout 👀✌️

src/utils/api/helper.ts
```diff
@@ -1054,6 +1063,98 @@ export const updateWallet = async (
...
+    EVENT_ID: TRANSACTION_AUTH_INIT.value,
+  });
+  await pipe(
+    TE.tryCatch(
+      () =>
+        apiPaymentEcommerceClient.requestTransactionAuthorization({
+          transactionId: checkData.idPayment,
+          body: {
+            amount: checkData.amount as any,
+            fee: getPspSelected().fee as any,
```
giovanniberti, line 1092: `as any` type assertions are not necessary. You can (and should) use `$TYPE.decode` for autogenerated types to convert between arbitrary objects and typesafe objects that use io-ts types

src/utils/api/helper.ts
```diff
@@ -1054,6 +1063,98 @@ export const updateWallet = async (
...
+                // TODO Qui ci va una error url?
+                return "fakePayment";
+              }
+            }
+          )
+        );
+        if (authorizationUrl !== "fakePayment") {
+          onResponse(authorizationUrl);
+        }
+      }
```
giovanniberti, line 1152: `TaskEither.fold` functions require a return type `Task<$STUFF>` (see [here](https://gcanti.github.io/fp-ts/modules/TaskEither.ts.html#fold)), while here you're returning `void`. Combined with the previous comment, you can rewrite this chain as:

```typescript
  const transactionId = checkData.idPayment;
  const authParams = {
    amount: checkData.amount,
    fee: getPspSelected().fee,
    paymentInstrumentId: getPaymentMethod().paymentMethodId,
    pspId: getPspSelected().pspCode,
    details:
      getPaymentMethod().paymentTypeCode === "CP"
        ? {
            detailType: "card",
            accountEmail: getEmailInfo(false).email,
            cvv: cardData.cvv,
            pan: cardData.pan,
            expiryDate: cardData.expiryDate,
            holderName: cardData.holderName,
          }
        : {
            detailType: "postepay",
            accountEmail: getEmailInfo(false).email,
          },
    language: LanguageEnum.IT,
  };

  await pipe(
    authParams,
    RequestAuthorizationRequest.decode,
    TE.fromEither,
    TE.chain(
      (request) => () =>
        apiPaymentEcommerceClient.requestTransactionAuthorization({
          transactionId,
          body: request,
        })
    ),
    TE.fold(
      (_r) => async () => {
        onError(ErrorsType.SERVER);
        mixpanel.track(TRANSACTION_AUTH_NET_ERROR.value, {
          EVENT_ID: TRANSACTION_AUTH_NET_ERROR.value,
        });
      }, // to be replaced with logic to handle failures
      (response) => {
        if (response.status === 200) {
          mixpanel.track(TRANSACTION_AUTH_SUCCES.value, {
            EVENT_ID: TRANSACTION_AUTH_SUCCES.value,
          });
          const authorizationUrl = response.value.authorizationUrl;
          onResponse(authorizationUrl);
          // eslint-disable-next-line @typescript-eslint/no-empty-function
          return T.fromIO(() => {});
        } else {
          onError(ErrorsType.GENERIC_ERROR);
          mixpanel.track(TRANSACTION_AUTH_RESP_ERROR.value, {
            EVENT_ID: TRANSACTION_AUTH_RESP_ERROR.value,
          });
          // TODO Qui ci va una error url?
          // eslint-disable-next-line @typescript-eslint/no-empty-function
          return T.fromIO(() => {});
        }
      }
    )
  )();
```

From a functional design point of view, callbacks such as `onError` and `onResponse` are an anti-pattern, and should be handled with proper types.
In this case, you're modeling a function that:
* Can return successfully with a `string` or fail and produce a `string` error value  => something like `Either<string, string>`
* Is async => in the case of `fp-ts` something like `Task`
So the right type here would be `TaskEither` (which you're already using, so why not return it? 😃)

Instead of using callbacks a better design would be to not use `fold` (because fold destroys the `Either` part of `TaskEither`), use `TE.bimap` and let downstream consumers handle the rest (i.e. error handling)

PR #177, opened 2023-02-07
[CHK-923] expiry date: use date formatter

ciuffagianluca:
Use a date formatter for parsing and writing card expiry date

#### List of Changes

Added date parser and formatter in check payment page

#### Types of changes

#### Checklist:


.env.development
```diff
@@ -9,6 +9,6 @@ CHECKOUT_POLLING_ACTIVATION_ATTEMPTS=3
 CHECKOUT_ENV=develop
 CHECKOUT_RECAPTCHA_SITE_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
 CHECKOUT_DONATIONS_URL=https://api.uat.platform.pagopa.it/donations/api/v1/availabledonations
-CHECKOUT_ECOMMERCE_HOST=http://localhost:1234
+CHECKOUT_ECOMMERCE_HOST=http://localhost:8080
```
giovanniberti, line 12: Doesn't this clash with CORS?

src/routes/PaymentCheckPage.tsx
```diff
@@ -123,8 +123,9 @@ export default function PaymentCheckPage() {
     window.location.replace(authorizationUrl);
   };
 
-  // TODO CHK-923
-  const [month, year] = cardData.expDate.split("/");
+  const stDate = "01/".concat(cardData?.expDate || "");
```
giovanniberti, line 126: What does `stDate` stands for? "Standardized date"? "String date"?

.env.development
ciuffagianluca: Need to revert it. Use it for local testing purpose.

src/routes/PaymentCheckPage.tsx
ciuffagianluca: It's a typo. string date was its original name.

src/routes/PaymentCheckPage.tsx
```diff
@@ -123,8 +123,9 @@ export default function PaymentCheckPage() {
...
+  const pattern = /(\d{2})\/(\d{2})\/(\d{2})/;
+  const dateParsed = new Date(stDate.replace(pattern, "$2-$1-$3"));
```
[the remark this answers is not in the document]
ciuffagianluca: This is a very elegant solution but it seems that our technology stack is a little bit old to support this notation.
Target in ts config need to be set to es6 and at runtime i get this error T`ypeError: String.prototype.matchAll called with a non-global RegExp argument`. Any suggestion?

.env.development
ciuffagianluca: Anyway this is .env.development

src/routes/PaymentCheckPage.tsx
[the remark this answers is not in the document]
giovanniberti: It seems that `matchAll` requires the `global` flag (else it throws a `TypeError`, see [here](https://stackoverflow.com/questions/60289995/matchall-throws-error-when-g-flag-is-missing-now) and, obviously missed by me, [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll) 😅.
In our case we expect one match only, so we can either add the `g` flag or use just [`match`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match) instead (without `g`, see docs)

```typescript
const [date, month, year] = stDate.match(regex).splice(1)
```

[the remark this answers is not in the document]
ciuffagianluca: Set in this way with to edit:

```
const datePattern = /(\d{2})\/(\d{2})\/(\d{2})/;
const [day, month, year] = stDate.match(regex).slice(1);
const dateParsed = new Date(Number(year)+2000, Number(month) - 1, Number(day));
```

first edit `slice` method in place of splice since it effectively return an array as expected
second edit in date constructor add 2000 to the year, since from documentation [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date) `year` parameter in date constructor is _Integer value representing the year. **Values from 0 to 99 map to the years 1900 to 1999**. All other values are the actual year.

src/routes/PaymentCheckPage.tsx
```diff
@@ -134,19 +134,23 @@ export default function PaymentCheckPage() {
     window.location.replace(authorizationUrl);
   };
 
-  const stDate = "01/".concat(cardData?.expDate || "");
-  const pattern = /(\d{2})\/(\d{2})\/(\d{2})/;
-  const dateParsed = new Date(stDate.replace(pattern, "$2-$1-$3"));
+  const dateString = "01/".concat(cardData?.expDate || "");
+  const datePattern = /(\d{2})\/(\d{2})\/(\d{2})/;
+  const [day, month, year] = (dateString.match(datePattern) || []).slice(1);
+  const dateParsed = new Date(Number(year)+2000, Number(month) - 1, Number(day));
```
giovanniberti, line 140: You could make it a little more safe (I haven't checked the surrounding context to see how the date is used btw)

```suggestion
  const dateString = pipe(
      O.fromNullable(cardData?.expDate),
      O.map(d => "01/".concat(d))
  );
  const datePattern = /(\d{2})\/(\d{2})\/(\d{2})/;

  const date = pipe(
      dateString,
      O.flatMap(date => {
          return O.fromNullable(date.match(datePattern))
      }),
      O.map(matches => matches.slice(1)),
      O.map(matches => [matches[0], matches[1], matches[2]]),
      O.map(([day, month, year]) => new Date(Number(year) + 2000, Number(month) - 1, Number(day))),
      O.getOrElse(null) // Only if there is no other choice :)
  );
```

src/utils/api/helper.ts
```diff
@@ -1357,6 +1357,42 @@ export const getCarts = async (
...
+    O.map((matches) => [matches[0], matches[1], matches[2]]),
+    O.map(
+      ([day, month, year]) =>
+        new Date(Number(year) + 2000, Number(month) - 1, Number(day))
+    ),
+    O.map((date) => expDateToString(date)),
+    O.fold(
+      () => "",
+      (parsedDate) => parsedDate
+    )
```
giovanniberti, line 1378:
```suggestion
    O.getOrElse("")
```

Even better, I would make this function return a `Option<string>` and let its users pipe into `getOrElse`

ciuffagianluca: Resolved making this function return a `Option<string>` and used into pipe with `getOrElse`

giovanniberti (approved):
🚀

PR #181, opened 2023-02-16
[#CHK-1111] Drop returnUrls from SessionStorage

vito80ba:
Use sessionStorage objects only when strictly necessary...

#### List of Changes
- definition of returnUrls' Interface without export, cause we use Cart instead
- drop returnUrls interface from all the files occurred
- refactoring of returnUrls with Cart.returnUrls when needed
- drop of ClearSensitiveItems and creation of ClearStorage (useful to reset sessionStorage dataset)

#### Motivation and Context

We don't need a specific key in sessionStorage to store returnUrls cause when a flow is started from the first page of Checkout every finally CTA (success, ko, cancel, etc) must redirect user to the home. When the flow starts with a Cart we can catch the urls from a subset of info called "returnUrls".

- Manual
- Linter
- E2e

#### Types of changes

### DIFF
--- a/src/features/payment/models/paymentModel.ts
+++ b/src/features/payment/models/paymentModel.ts
@@ -172,7 +172,7 @@ export interface PaymentNotice {
   description?: string;
 }
 
-export interface ReturnUrls {
+interface ReturnUrls {
   returnOkUrl: string;
   returnCancelUrl: string;
   returnErrorUrl: string;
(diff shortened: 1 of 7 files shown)


ciuffagianluca (approved):
Well done!

stratoivandiluccio (approved):

PR #183, opened 2023-02-20
Base64 transaction id decode

Fernando-Granato:
#### List of Changes
Added method to decode Base64 transactionId received after PGS authorization

#### Motivation and Context
Decoded the transactionId received after PGS authorization to poll the transaction-service

#### Types of changes

#### Checklist:


ciuffagianluca (approved):
LGTM ✌️

src/utils/api/response.ts
```diff
@@ -29,6 +29,8 @@ const config = getConfigOrThrow();
 const retries: number = 10;
 const delay: number = 3000;
 const timeout: Millisecond = config.CHECKOUT_API_TIMEOUT as Millisecond;
+const decode = (transactionId: string): string => Buffer.from(transactionId, 'base64').toString('binary');
+
```
pietro-tota, line 33: The base64 string sent from transactions-service is calculated starting from the bytes of the UUID (high and low bytes) so here you have to build UUID structure again as shared [here](https://playcode.io/1195321)

pietro-tota (changes requested):
Seems that the UUID conversion does not build UUID correctly, see comment

src/utils/api/response.ts
Fernando-Granato: You are absolutely right

Fernando-Granato: @pietro-tota  What do you think now 😉

pietro-tota (approved):
Well done!🚀

PR #192, opened 2023-03-07
Chk 1207 psp disclaimer

ciuffagianluca:
This PR add a PSP Disclaimer based on the amount and removes it in case the user updates the psp based in his preference

#### List of Changes
- Added psp disclaimer
- Added translation for disclaimer
- Added some unit test for the moneyFormat utility function
- Threshold management from gec response

#### Motivation and Context
give to the user a feedback about the default selected psp and the reason

manually using as BE the DEV environment 

#### Types of changes

#### Checklist:


stratoivandiluccio (approved):

src/routes/InputCardPage.tsx
```diff
@@ -91,8 +91,8 @@ export default function InputCardPage() {
           BundleOption.decode,
           O.fromEither,
           O.chain((r) => O.fromNullable(r.belowThreshold)),
-          O.fold(
-            () => undefined,
+          O.foldW(
+            () => onError,
```
stratoivandiluccio, line 95: Please note that you are defining an anonymous function that when executed will not call the onError function

vito80ba (approved):

PR #200, opened 2023-03-23
[CHK-1298] Disabling redux devTools in PROD

stratoivandiluccio:
#### List of Changes
The redux devTools will be not available in production.

#### Motivation and Context
It's much more safer to does not expose via redux devTools the internal FE state. 
Actions/events that could alter the FE behaviour could be dispatched when redux devTools are enabled.

Tested locally updating the CHECKOUT_ENV environment variable with 'DEV' | 'UAT' | 'PROD' values.
Then I verified in the browser if the redux devTools were enabled or not.

#### Types of changes

#### Checklist:


vito80ba (approved):
Great proposal!

PR #226, opened 2023-06-15
[CHK-1481] fix: update carts amount

MarcoAGreco:
#### List of Changes

- Update cart notices amount after activation

#### Motivation and Context

- Avoid amount inconsistency between getCarts and payment activation
- build, unit

#### Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

#### Checklist:


src/utils/api/helper.ts
```diff
@@ -262,6 +262,29 @@ const activePaymentTask = (
...
+                const cartInfo = getSessionItem(SessionItems.cart) as
+                  | Cart
+                  | undefined;
+                if (cartInfo !== undefined) {
+                  const updatedPaymentNotices = cartInfo.paymentNotices.map(
+                    (paymentNotice) => {
+                      const updatedAmount = responseType.value.payments.find(
+                        (p) =>
+                          p.rptId ===
+                          (`${paymentNotice.fiscalCode}${paymentNotice.noticeNumber}` as RptId)
```
giovanniberti, line 274: Is this cast ok? Isn't `RptId` an `io-ts` tagged type?

src/utils/api/helper.ts
```diff
@@ -262,6 +262,29 @@ const activePaymentTask = (
...
               if (responseType.status === 200) {
                 reason = "";
+                const cartInfo = getSessionItem(SessionItems.cart) as
+                  | Cart
+                  | undefined;
+                if (cartInfo !== undefined) {
+                  const updatedPaymentNotices = cartInfo.paymentNotices.map(
+                    (paymentNotice) => {
+                      const updatedAmount = responseType.value.payments.find(
+                        (p) =>
```
giovanniberti, line 272: [**suggestion (if-minor)**:](https://conventionalcomments.org/) _map_ + _find_ makes this an $O(n^2)$ algorithm, what if we first map the actualized amounts from the activate response into a `RptId -> amount` map and then update the amounts?

MarcoAGreco: You're right! Solved with 3c936a5 😁

src/utils/api/helper.ts
MarcoAGreco: I hope it is, we use it several times in this PR

giovanniberti: Just checked, _as a type_ `RptId` is an alias of `string & PatternedString<Pattern>`, so can be safely coerced with `as`

giovanniberti (approved):
🚀

PR #228, opened 2023-07-05
[CHK-1673] Update drawer

MarcoAGreco:
#### List of Changes
- Update side drawer with cart icon + question mark icon (mui: `HelpOutlineOutlined`)

#### Types of changes

#### Checklist:


Fernando-Granato (approved):
🚀

src/components/commons/Header.tsx
```diff
@@ -104,7 +105,13 @@ export default function Header() {
                   onClick={() => toggleDrawer(true)}
                 >
                   {moneyFormat(amountToShow())}
-                  <InfoOutlinedIcon color="primary" sx={{ ml: 1 }} />
+                  <img
+                    src={cartIcon}
+                    alt="Icona carrello"
```
Fernando-Granato, line 110: is the alt text right in Italian language ?

src/components/commons/Header.tsx
```diff
@@ -18,6 +18,7 @@ import { getTotalFromCart } from "../../utils/cart/cart";
 import { moneyFormat } from "../../utils/form/formatters";
 import { paymentSubjectTransform } from "../../utils/transformers/paymentTransformers";
 import DrawerDetail from "../Header/DrawerDetail";
+import cartIcon from "../../assets/images/cart.png";
```
stratoivandiluccio, line 22: It's much better to use ShoppingCartIcon component exported from '@mui/icons-material' instead of using a low resolution raster png. Plus, the problem about the alt property will not exist anymore
https://mui.com/material-ui/material-icons/?query=ca&selected=ShoppingCart

Fernando-Granato: I do not think it is possible to use the icon provided by material-icons @infantesimone @MarcoAGreco confirmed ?

src/components/commons/Header.tsx
MarcoAGreco: I've assumed that it's ok, you can find other imgs with the same approach. Like [here](https://github.com/pagopa/pagopa-checkout-fe/blob/aa69616a71fbb225fbfdf3e16f684033d557996f/src/components/TextFormField/PspFieldContainer.tsx#L56)

infantesimone (approved):

PR #231, opened 2023-07-14
[CHK-1690] Qualtrics survey on thank you page

stratoivandiluccio:
#### List of Changes

Added an info alert component which includes a link to a Qualtrics survey on the thank you page (alert visible only when the payment is ok)

#### Motivation and Context
The purpose is to measure the ease of the payment process experience

Tested locally using the be-mock as backend and simulating an ok and ko payment

image

#### Types of changes

#### Checklist:

### DIFF
--- a/src/routes/PaymentResponsePage.tsx
+++ b/src/routes/PaymentResponsePage.tsx
@@ -92,6 +92,7 @@ export default function PaymentResponsePage() {
         idStatus,
         outcome,
       });
+      setOutcome(outcome);
       showFinalResult(outcome);
     };
 
@@ -105,7 +106,6 @@ export default function PaymentResponsePage() {
           : cart
           ? cart.returnUrls.returnErrorUrl
           : "/";
-      setOutcome(outcome);
       setOutcomeMessage(message);
       setRedirectUrl(redirectTo || "");
       setLoading(false);
@@ -176,7 +176,7 @@ export default function PaymentResponsePage() {
               </Button>
             </Box>
             {outcome === ViewOutcomeEnum.SUCCESS && (
-              <Box>
+              <Box sx={{ width: "100%" }}>
                 <SurveyLink />
               </Box>
             )}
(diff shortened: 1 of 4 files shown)


src/routes/PaymentResponsePage.tsx
```diff
@@ -172,6 +175,11 @@ export default function PaymentResponsePage() {
                 {t("errorButton.close")}
               </Button>
             </Box>
+            {outcome === ViewOutcomeEnum.SUCCESS && (
+              <Box sx={{ width: "100%" }}>
```
vito80ba, line 179: Discuss this with the designer showing the draft you made.
Maybe he'll prefer a width inlined with the previous element (only on desktop viewport I guess).
Schermata 2023-07-14 alle 12 59 26

src/translations/it/translations.ts
```diff
@@ -312,6 +312,14 @@ export const TRANSLATIONS_IT = {
       title: "L'operazione è stata presa in carico",
       body: "Riceverai l’esito del pagamento all'indirizzo {{useremail}}",
     },
+    survey: {
```
vito80ba, line 315: Suggestion: isn't better to _name_ this with a more specific label?
Try to imagine a situation where you've more surveys in the webapp... how manage this in a structured way?

src/translations/it/translations.ts
```diff
@@ -312,6 +312,14 @@ export const TRANSLATIONS_IT = {
...
+      title: "Puoi dirci com'è andata?",
+      body: "Raccontaci la tua esperienza con il pagamento e aiutaci a migliorare.",
+      link: {
+        text: "Vai al sondaggio",
+        href: "https://pagopa.gov.it/diccilatua/ces-pagamento",
```
vito80ba, line 320: Add a trail slash... it's more safe
`https://pagopa.gov.it/diccilatua/ces-pagamento/`

src/translations/it/translations.ts
stratoivandiluccio: I think it's not necessary because survey is a property of paymentResponsePage and we can distinguish several surveys on different pages like the example below
`paymentResponsePage.survey
 paymentCheckPage.survey`

vito80ba (approved):

giovanniberti (approved):

PR #233, opened 2023-07-19
[Chk-1719] aggiornare logo pagopa

ciuffagianluca:
Update pagopa-logo referred from notifications service to put it in email https://github.com/pagopa/pagopa-email-templates/tree/main/assets

#### List of Changes

Update pagopalogo using a png file compliant with gmail client

#### Motivation and Context

Pagopa logo wasn't visible in gmail client

#### Types of changes

#### Checklist:


Fernando-Granato (changes requested):
In this point the old logo image is used  to show it in the header https://github.com/pagopa/pagopa-checkout-fe/blob/ea53d9506a8e56ae2ea31037be4d84c934644da6/src/components/commons/Header.tsx#L6 but the new logo image have a different name

Fernando-Granato (changes requested):
In which case the white logo is used ? 🙂

ciuffagianluca:
> In this point the old logo image is used to show it in the header
> 
> https://github.com/pagopa/pagopa-checkout-fe/blob/ea53d9506a8e56ae2ea31037be4d84c934644da6/src/components/commons/Header.tsx#L6
> 
> but the new logo image have a different name



> In which case the white logo is used ? 🙂

By now it is not used, but it is part of the offical logo repo so I think it could be useful to import it

ciuffagianluca:
> In this point the old logo image is used to show it in the header
> 
> https://github.com/pagopa/pagopa-checkout-fe/blob/ea53d9506a8e56ae2ea31037be4d84c934644da6/src/components/commons/Header.tsx#L6
> 
> but the new logo image have a different name

This is another ref, to the image dir, while we are replacing the one in logos dir

ciuffagianluca:
> > > In this point the old logo image is used to show it in the header
> > > https://github.com/pagopa/pagopa-checkout-fe/blob/ea53d9506a8e56ae2ea31037be4d84c934644da6/src/components/commons/Header.tsx#L6
> > > 
> > > but the new logo image have a different name
> > 
> > 
> > This is another ref, to the image dir, while we are replacing the one in logos dir
> 
> Are we sure that we do not need to update the image in the assets/images folder as well ?

These are different resource that belongs to different domains. I agree with you that it should be better to have single logo repo and single logo ref, but changing it could lead to undesired frontend issue. Moreover, unify logos is out of the scope of this PR. This could be a good point to mrk as tech debt and report in a github issue. What do you think?

MarcoAGreco (approved):

Fernando-Granato (approved):

PR #252, opened 2023-08-31
[CHK-1826] fix card form duplication

mmoio:
#### List of Changes

- payment choice page uses location assign to navigate to card data page 
- payment check page removes `beforeunload` listener before navigating back to card data page
- payment check page uses location replace to navigate back to card data page 

#### Motivation and Context

Use location replace or assign instead of router-dom navigate so to
cause a window reload and cleaning the window from frames
loaded by previous npg sdk builds

Visual testing

#### Types of changes

#### Checklist:

### DIFF
--- a/src/features/payment/components/PaymentChoice/PaymentChoice.tsx
+++ b/src/features/payment/components/PaymentChoice/PaymentChoice.tsx
@@ -2,7 +2,6 @@
 /* eslint-disable @typescript-eslint/no-unused-expressions */
 /* eslint-disable functional/immutable-data */
 import React from "react";
-import { useNavigate } from "react-router";
 import ClickableFieldContainer from "../../../../components/TextFormField/ClickableFieldContainer";
 import {
   PaymentMethodRoutes,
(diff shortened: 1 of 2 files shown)


ciuffagianluca:
Hi @mmoio ,
this solution solve the issue between the PaymentCheckPage and IFrameCardPage.
But the bug is still present between PaymentChoicePage and IFrameCardPage. In that case too the invocation is replicated n times if you navigate n times between the 2 pages.

ciuffagianluca (commented):
Hi @mmoio ,
this solution solve the issue between the PaymentCheckPage and IFrameCardPage.
But the bug is still present between PaymentChoicePage and IFrameCardPage. In that case too the invocation is replicated n times if you navigate n times between the 2 pages.

mmoio:
> Hi @mmoio , this solution solve the issue between the PaymentCheckPage and IFrameCardPage. But the bug is still present between PaymentChoicePage and IFrameCardPage. In that case too the invocation is replicated n times if you navigate n times between the 2 pages.

Hi, I've updated the PaymentChoicePage to use location for navigation to IFrameCardPage

ciuffagianluca (approved):
Well done! 🚀

vito80ba (approved):

PR #253, opened 2023-09-04
feat(jwt): CHK-1860 jwt check into retrieve card data method

Fernando-Granato:
#### List of Changes
Added bearerAuth and x-transaction-id-from-client header into call for retrieve card data.

#### Motivation and Context
The change was made to securitize the api for card data retrieval.

#### Types of changes

#### Checklist:

### DIFF
--- a/src/utils/api/helper.ts
+++ b/src/utils/api/helper.ts
@@ -380,7 +380,6 @@ export const retrieveCardData = async ({
   onError: (e: string) => void;
   onResponseSessionPaymentMethod: (r: any) => void;
 }) => {
-  /*
   const transactionId = pipe(
     getSessionItem(SessionItems.transaction) as NewTransactionResponse,
     O.fromNullable,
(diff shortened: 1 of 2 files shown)


src/utils/api/helper.ts
```diff
@@ -398,13 +397,15 @@ export const retrieveCardData = async ({
   mixpanel.track(PAYMENT_PSPLIST_INIT.value, {
```
ciuffagianluca, line 397: PR looks great to me. But maybe in this point we could delete/comment this mixpanel invocation?

Fernando-Granato: We should eliminate all events in that method what do you think ?

ciuffagianluca: Just this event, because the psp list initialization start one api call down the stream, not precisely here in this point

Fernando-Granato: Fixed with 18445e92a68a78ef97f56923ba9781c1d3103e0b

ciuffagianluca (approved):
LGTM! 🚀

infantesimone (approved):

PR #254, opened 2023-09-06
[#CHK-1866] new detail to post for auth-request

vito80ba:
The aim of this task/PR is to drop cardData from the auth-request post. The new arch works only with the sessionId provided in the previous call (via NPG)

#### List of Changes

- Drop cardData 
- Added sessionId to the post
- Drop unuseful imports

#### Motivation and Context

The aim of this task/PR is to drop cardData from the auth-request post. The new arch works only with the sessionId provided in the previous call (via NPG)

#### Types of changes

### DIFF
--- a/src/utils/api/helper.ts
+++ b/src/utils/api/helper.ts
@@ -589,11 +589,7 @@ export const calculateFees = async ({
 };
 
 export const proceedToPayment = async (
-  {
-    transaction,
-  }: {
-    transaction: NewTransactionResponse;
-  },
+  transaction: NewTransactionResponse,
   onError: (e: string) => void,
   onResponse: (authUrl: string) => void
 ) => {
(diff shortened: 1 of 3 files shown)


src/utils/api/helper.ts
```diff
@@ -592,16 +591,8 @@ export const calculateFees = async ({
...
   }: {
     transaction: NewTransactionResponse;
-    cardData: {
-      brand: string;
-      cvv: string;
-      pan: string;
-      expiryDate: string;
-      holderName: string;
-    };
   },
```
stratoivandiluccio, line 596: Here, I think it's better to have just one 'transaction' param instead of an object with only one 'transaction' property.

src/utils/api/helper.ts
```diff
@@ -655,7 +647,6 @@ export const proceedToPayment = async (
       (getSessionItem(SessionItems.useremail) as string) || "",
     mobilePhone: null,
   };
-
```
stratoivandiluccio, line 658: For consistency  I would restore this space

stratoivandiluccio (approved):

PR #256, opened 2023-09-07
[CHK-1867] New /gdi-check route added

stratoivandiluccio:
#### List of Changes

Added a new /gdi-check route page. It contains a spinner and an hidden iframe, which src is obtained from the fragment url. The route is protected by a guard on the sessionId

#### Motivation and Context

The new /gdi-check route is a necessary step of the payment flow

Tested locally, removing temporary the guard, and using a generic url for the hidden iframe

#### Types of changes

#### Checklist:

- [x] My change requires a change to the documentation.


src/App.tsx
```diff
@@ -158,9 +159,13 @@ export function App() {
               />
               <Route
                 path={CheckoutRoutes.RIEPILOGO_PAGAMENTO}
```
infantesimone, line 161: Why delete the guard on RIEPILOGO_PAGAMENTO?

src/utils/__tests__/urlUtili.test.ts
```diff
@@ -0,0 +1,34 @@
+import { getFragmentParameter } from "../../utils/regex/urlUtilities";
```
infantesimone, line 1: great!

vito80ba (approved):
LGTM

src/App.tsx
stratoivandiluccio: Maybe a mistake. I see that there was two guards:
image
I think only the one on the SessionItems.transaction needs to be restore, because we are not managing cvv data FE side anymore. Do you agree?

infantesimone: yep :-)

PR #264, opened 2023-09-19
fix(iFrameCard): using npg error to set focus

mmoio:
#### List of Changes

- Uses npg sdk error and flowChange events to set focus
- Removes the FocusProxy component to intercept focus

#### Motivation and Context

The focus proxy approach does not work on safari, this pr attempts to fix this issue. 
This approach causes a delay in the state update, since the events are sent on the first key press in an input.

Visual testing

#### Types of changes

#### Checklist:


stratoivandiluccio (approved):

vito80ba (approved):
LGTM

PR #274, opened 2023-10-06
Theme customization

AnisLucidi01:
#### List of Changes

Changed theme font, customized color and buttons to make it more similar to IO's UI.

#### Motivation and Context

As stated above, we need to make our theme similar to IO's to make it so that the user won't notice the switch from IO App to our webview. 

This has been also tested for mobile interface. 

image
#### Types of changes

#### Checklist:


src/App.tsx
```diff
@@ -4,6 +4,8 @@ import { theme } from "@pagopa/mui-italia";
 import React from "react";
 import { useTranslation } from "react-i18next";
 import { BrowserRouter, Navigate, Route, Routes } from "react-router-dom";
+import type { MuiLoadingButton } from '@mui/lab/themeAugmentation';
```
mmoio, line 7: To avoid the typescript error you can change this line to:
import "@mui/lab/themeAugmentation";

src/App.tsx
```diff
@@ -63,6 +84,37 @@ const checkoutTheme = createTheme({
...
+      styleOverrides: {
+        root: {
+          borderRadius: "8px",
+        },
+      },
+    },
+    MuiLoadingButton: {
+      styleOverrides: {
+        loadingIndicator: {
+          color: "white !important",
```
mmoio, line 97: This !important can be removed, also the loadingIndicator should have a size of 24px

src/App.tsx
```diff
@@ -63,6 +84,37 @@ const checkoutTheme = createTheme({
...
+        },
+        root: ({ ownerState }) =>
```
mmoio, line 99: You can avoid using important and the conditionals with "&.MuiLoadingButton-loading" as a subkey of root instead of using the ownerState.

src/App.tsx
```diff
@@ -63,6 +84,37 @@ const checkoutTheme = createTheme({
...
+          (ownerState.loading && {
+            backgroundColor: "#0B3EE3 !important",
```
mmoio, line 101: To avoid repeating the hex color string, you can use the theme argument provided in the root callback.

src/App.tsx
```diff
@@ -63,6 +84,37 @@ const checkoutTheme = createTheme({
...
+            color: "transparent !important",
+          }) ||
+          (ownerState.disabled &&
```
mmoio, line 104: This should be taken up to the designers, but for this pr we can assume the button must stay grayed out when disabled, so you can just remove the (103-107) lines

mmoio (changes requested):

mmoio:
FYI the way you added the fonts is fine, but i think you could just use @fontsource/readex-pro, since is the approach used by mui-italia for Titillium

src/App.tsx
mmoio: Also it seems that the button text should not be transformed to capitalized anymore

src/App.tsx
AnisLucidi01: Thanks for the useful info! I was trying to understand how to do it.

src/App.tsx
```diff
@@ -94,18 +82,16 @@ const checkoutTheme = createTheme({
     MuiLoadingButton: {
       styleOverrides: {
         loadingIndicator: {
-          color: "white !important",
+          color: "white",
+          size: "24px",
```
mmoio, line 86: Size appear to not be a valid properties for this element, you could try setting width and height

src/App.tsx
```diff
@@ -94,18 +82,16 @@ const checkoutTheme = createTheme({
...
-          }) ||
-          (ownerState.disabled &&
-            ownerState.type === "submit" && {
-              backgroundColor: "#0B3EE3 !important",
-              color: "white !important",
-            }),
+        root: ({ theme }) => ({
+          textTransform: "capitalize",
+          "&.MuiLoadingButton-loading": {
+            backgroundColor: theme.palette.primary,
```
mmoio, line 91: Should be `theme.palette.primary.main`

mmoio (changes requested):

mmoio (approved):

vito80ba:
Well done @AnisLucidi01 , but for now this PR couldn't be merged

PR #279, opened 2023-10-13
[CHK-1957] Auth Request error handling

mmoio:
#### List of Changes
- Catches and handles error in the Auth Request api call
- Shows error popup and enables submit button in case of errors

#### Motivation and Context

Previously errors in the Auth Request api call were not handled correctly, causing the user to be stuck on the check/payment page in case of errors, with a disabled button and no way to retry the payment call

#### Types of changes

#### Checklist:


mmoio:
Leaving in draft since is not possible to test the changes when the auth request is successful, since is responding with a 504 status code.

vito80ba (commented):
LGTM

PR #285, opened 2023-11-29
[#CHK-2178] Survey as feature flag

vito80ba:
We want to manage the final survey with a feature flag (disabling it, for now...)

#### List of Changes

- added specific env to the config list
- mod the condition to display it in the response page

#### Motivation and Context

For now, we wan't to display the final survey. Maybe in the next future

#### Types of changes

#### Checklist:


src/routes/PaymentResponsePage.tsx
```diff
@@ -175,11 +177,12 @@ export default function PaymentResponsePage() {
                 {t("errorButton.close")}
               </Button>
             </Box>
-            {outcome === ViewOutcomeEnum.SUCCESS && (
-              <Box sx={{ width: "100%" }} px={{ xs: 8, sm: 0 }}>
-                <SurveyLink />
-              </Box>
-            )}
+            {conf.CHECKOUT_SURVEY_SHOW === "true" &&
+              outcome === ViewOutcomeEnum.SUCCESS && (
```
stratoivandiluccio, line 181: Can we avoid to use the string "true" or "false" if it's not possible to use a boolean type? Something like "show" or "hidden"

.devops/pagopa-deploy-pipelines.yml
```diff
@@ -351,6 +353,7 @@ stages:
               api_ecommerce_basepath: '/ecommerce/checkout/v1'
               pagopa_assets_cdn: 'https://assets.cdn.platform.pagopa.it/payment-methods'
               pagopa_logos_cdn: 'https://assets.cdn.io.italia.it/logos/abi/'
+              survey_show: 'false'
```
stratoivandiluccio, line 356: I think should be 0 or 1

stratoivandiluccio (approved):

PR #295, opened 2023-12-11
Fix: [CHK-2227] await npg sdk before build

mmoio:
#### Types of changes

#### Checklist:


vito80ba (approved):

src/hooks/useNpgSdk.ts
```diff
@@ -0,0 +1,60 @@
...
+}: SdkBuild) => {
+  const [sdkReady, setSdkReady] = useState(false);
+
+  const createBuild = () => {
+    try {
+      // THIS is mandatory cause the Build class is defined in the external library called NPG SDK
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      const newBuild = new Build(
+        createBuildConfig({
```
stratoivandiluccio, line 36: newBuild is declared but never used, I think we can just instantiate. (just new Build)

src/hooks/useNpgSdk.ts
```diff
@@ -0,0 +1,60 @@
...
+  onChange?: (field: FieldId, fieldStatus: FieldStatus) => void;
+  onReadyForPayment?: () => void;
+  onPaymentComplete?: () => void;
+  onPaymentRedirect?: (urlRedirect: string) => void;
+  onBuildError: () => void;
+};
+
+const noop = () => {
+  // noop
+};
```
stratoivandiluccio, line 19: what's the purpose of this empty function?

src/hooks/useNpgSdk.ts
```diff
@@ -0,0 +1,60 @@
...
+  onPaymentRedirect = () => null,
+  onBuildError,
+}: SdkBuild) => {
+  const [sdkReady, setSdkReady] = useState(false);
+
+  const createBuild = () => {
+    try {
+      // THIS is mandatory cause the Build class is defined in the external library called NPG SDK
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
stratoivandiluccio, line 34: I think it's time to get rid of this @ts-ignore globally. To do that try to rename the file image.d.ts in modules.d.ts (or maybe globals.d.ts) ad append these lines.
`/** Build class defined in the external NPG SDK library */`
`declare const Build: any;`

src/hooks/useNpgSdk.ts
mmoio: I'm using this to initialize the sdkReady state as an empty function, so that the returned function can be called without errors even if the sdk build is not ready. I'could change this behavior and make the function throw an error if the sdk is not ready.

src/hooks/useNpgSdk.ts
mmoio: Removed the unused declaration

src/hooks/useNpgSdk.ts
mmoio: I've followed your suggestion and was able to drop the rule-disable comments

stratoivandiluccio (approved):

PR #347, opened 2024-05-16
[CHK-2851] fix exclude path double path

ciuffagianluca:
Fix deploy-pipeline

#### List of Changes

The double `exclude-path` option for sync command exclude the first one and undesired delete happens

#### Types of changes

#### Checklist:

### DIFF
--- a/.devops/pagopa-deploy-pipelines.yml
+++ b/.devops/pagopa-deploy-pipelines.yml
@@ -117,7 +117,7 @@ stages:
               azureSubscription: 'DEV-PAGOPA-SERVICE-CONN'
               scriptLocation: inlineScript
               inlineScript: |
-                call az storage blob sync --container $web --account-name pagopadcheckoutsa -s "$(Pipeline.Workspace)\Bundle_DEV" --exclude-path payment-transactions-gateway --exclude-path ecommerce-fe
+                call az storage blob sync --container $web --account-name pagopadcheckoutsa -s "$(Pipeline.Workspace)\Bundle_DEV" --exclude-path payment-transactions-gateway ecommerce-fe
 
           - task: AzureCLI@1
             displayName: 'Purge CDN endpoint on DEV'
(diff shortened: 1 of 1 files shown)


vito80ba (approved):

.devops/pagopa-deploy-pipelines.yml
```diff
@@ -117,7 +117,7 @@ stages:
               azureSubscription: 'DEV-PAGOPA-SERVICE-CONN'
               scriptLocation: inlineScript
               inlineScript: |
-                call az storage blob sync --container $web --account-name pagopadcheckoutsa -s "$(Pipeline.Workspace)\Bundle_DEV" --exclude-path payment-transactions-gateway --exclude-path ecommerce-fe
+                call az storage blob sync --container $web --account-name pagopadcheckoutsa -s "$(Pipeline.Workspace)\Bundle_DEV" --exclude-path payment-transactions-gateway ecommerce-fe
```
pietro-tota, line 120:
```suggestion
                call az storage blob sync --container $web --account-name pagopadcheckoutsa -s "$(Pipeline.Workspace)\Bundle_DEV" --exclude-path payment-transactions-gateway;ecommerce-fe
```
looking at [azcopy docs](https://learn.microsoft.com/en-us/azure/storage/common/storage-ref-azcopy-copy) multiple paths have to be separated by semicolon

pietro-tota (changes requested):
azcopy --exclude-path paths should be separated by semicolon, right?

pietro-tota (approved):

.devops/pagopa-deploy-pipelines.yml
ciuffagianluca: Set in this way while you were posting :)

ciuffagianluca:
Bypass branch protection for integration test failed as known in main branch

PR #374, opened 2024-09-12
fix: [CHK-3225] Integration tests

petretiandrea:
#### List of Changes

- fix headless configuration
- update integration test suites
- fix onetrust configuration for test envs
- introduce language setting for integration tests

#### Motivation and Context

This PR has focus on integration test restoring after some months of constant failure

Tests in local and remote env

#### Types of changes

#### Checklist:


static/ot/consent/c4394b6a-86ce-428c-a3d3-5b81cabbfed4-test/c4394b6a-86ce-428c-a3d3-5b81cabbfed4-test.json
```diff
@@ -1 +1 @@
...
...tedTime":"2024-05-31T15:54:20.398736817","updatedTime":"2024-05-31T15:54:20.398746817","cmpId":"28","cmpVersion":"1","consentScreen":"1","consentLanguage":null,"vendorListVersion":0,"maxVendorId":0,"encodingType":"0","globalVendorListUrl":"https://cdn.cookielaw.org/vendorlist/iab2V2Data.json"},"GoogleData":{"vendorListVersion":1,"googleVendorListUrl":"https://cdn.cookielaw.org/vendorlist/googleData.json"},"ScriptDynamicLoadEnabled":false,"TenantFeatures":{"CookieV2BannerFocus":true,"CookieV2RejectAll":true,"CookieV2GPC":true,"CookieV2AssignTemplateRule":true,"MobileAuthenticatedConsents":true,"CookieV2GeolocationJsonApi":true,"CookieV2GCMDMA":true,"CookieV2TCF21":true,"CookieV2RemoveSettingsIcon":true,"CookieV2BannerLogo":true,"CookieV2GeneralVendors":true,"CookieV2GPP":true},"IsSuppressBanner":false,"IsSuppressPC":false,"PublisherCC":"DE","Domain":"checkout.pagopa.it","TenantGuid":"77f17844-04c3-4969-a11d-462ee77acbe1","EnvId":"app-de-prod","RemoteActionsEnabled":false,"GeoRuleGroupName":"Solo tecnici","GATrackToggle":true,"GATrackAssignedCategory":"AS","WebFormIntegrationEnabled":false,"WebFormSrcUrl":"","WebFormWorkerUrl":"","GppData":{"cmpId":"28"},"AuthenticatedConsent":false}
```
infantesimone, line 1: why this?

ciuffagianluca: It's necessary to make test successful.
One trust js file were imported only for prod env, but not in other envs. In other envs there was simply an nonexistent link that leads tests to fail. You can experiment this failure also in dev or uat env now, simply loading checkout portal and opening console. With this json, we provide an alternative root for onetrust files, but it is a fake configuration, that will make the cookie policy popup not appearing in test envs, but allow the tests to run successfully.

src/__integration_tests__/final-status-npg.integration.test.ts
```diff
@@ -15,31 +17,31 @@ describe("Unauthorized npg final status mapping tests", () => {
   const mockFlowWithExpectedResultMap = new Map([
     [
       "AUTHORIZATION_REQUESTED_NO_NPG_OUTCOME",
-      { title: "AUTHORIZATION REQUESTED WITH NO NPG OUTCOME", esito: "L'operazione è stata presa in carico" }
+      { title: "AUTHORIZATION REQUESTED WITH NO NPG OUTCOME", esito: "Attendi la conferma del pagamento" }
```
pietro-tota, line 20: what about take those description from the expected translation file key property?

something like
```code
import italianTranslations from "../translations/it/translations.json";

//and the test below will become
[
      "AUTHORIZATION_COMPLETED_WITH_NPG_AUTH_STATUS_EXECUTED",
      { title: "AUTHORIZATION COMPLETED WITH NPG AUTHORIZATION STATUS EXECUTED", esito: italianTranslations["GENERIC_ERROR"].title }
    ],
```

This way the test will continue to work properly even in the case of future translations update and here the correct outcome - translation key to be shown will be tested instead of performing an exact text match that will require all tests to be updated in case of any modification

ciuffagianluca: It's not so simple as it seems to be. Anyway this is a good idea, but to be handled in another PR.

infantesimone (approved):

src/__integration_tests__/final-status-npg.integration.test.ts
pietro-tota: Yes, agree that it can be done into another PR :) 

> It's not so simple as it seems to be. 

I've tested this piece of code before suggesting it successfully, replacing all the `esito: "Si è verificato un errore imprevisto"` with `esito: italianTranslations["GENERIC_ERROR"].title`and executing tests checking for correct execution. 
Tested for just one case test it become a simple "replace all" operation for all the remaining string instances hardcoded into tests so it seems fair simple to me.
Do you see any other complication that I'm missing here?

giovanniberti (approved):

PR #395, opened 2024-11-12
Bugfix [CHK-3540] PSP sorting on fee 0 is now correct

EmanueleBVtech:
depends on https://github.com/pagopa/pagopa-checkout-be-mock/pull/82

#### List of Changes
Solved minor bug causing bad sorting on PSP with fee = 0

#### Motivation and Context
Needed proper sorting on PSP Drawer

Manual test

#### Types of changes

#### Checklist:


src/features/payment/components/PaymentPspDrawer/PaymentPspDrawer.tsx
```diff
@@ -125,8 +125,11 @@ const sortBy =
     const fieldA = a[field];
     const fieldB = b[field];
     const order = direction === "asc" ? 1 : -1;
-
-    return fieldA && fieldB ? (fieldA > fieldB ? order : -order) : -order;
+    return fieldA !== undefined && fieldB !== undefined
```
scaminati-bv, line 128: With strict equality you miss "null" check. In this case is better to use simple one (`!=`) or add also "null" check

EmanueleBVtech: this was omitted as the related API 
`/payment-methods/{id}/fees` is granted by open API to return an object 
CalculateFeeResponse containing an array of Bundle who has the sorted attribute **taxPayerFee**.

The value **null** is granted to never happen as per OpenAPI signature. The undefined control is added only to avoid crashes in case the attribute is missing due to misconstructed payloads.

scaminati-bv: Ok if null value never happen undefined check is ok

src/features/payment/components/PaymentPspDrawer/PaymentPspDrawer.tsx
```diff
@@ -125,8 +125,11 @@ const sortBy =
...
+      ? fieldA > fieldB
```
pietro-tota, line 129: Please add tests for this function

EmanueleBVtech: done here 883526c55b558201a6cab035cff148246a3d10f6

.devops/pagopa-code-review-pipelines.yml
```diff
@@ -7,232 +7,228 @@
 #
 
 variables:
-  NODE_VERSION: '14.16.0'
+  NODE_VERSION: "14.16.0"
```
pietro-tota, line 10: it seems that there is no change here except for file spaces and apex, if so please restore this file

jest-puppeteer.config.js
```diff
@@ -1,9 +1,9 @@
...
-        headless: "new",
-        product: 'chrome',
-        args: ["--no-sandbox"] // to fix
-    },
-    browserContext: 'incognito'
-}
\ No newline at end of file
+  launch: {
+    dumpio: true,
+    headless: "false",
```
pietro-tota, line 4:
```suggestion
    headless: "new",
```

restore this value to headless new

EmanueleBVtech: done here 397425e1a47f87953fb0b7d2c0f4efc11e217034

src/features/payment/components/PaymentPspDrawer/PaymentPspDrawer.tsx
```diff
@@ -120,15 +125,6 @@ export const PaymentPspDrawer = (props: {
...
-const sortBy =
-  (field: PspField, direction: "asc" | "desc") => (a: Bundle, b: Bundle) => {
-    const fieldA = a[field];
-    const fieldB = b[field];
-    const order = direction === "asc" ? 1 : -1;
-
-    return fieldA && fieldB ? (fieldA > fieldB ? order : -order) : -order;
-  };
-
 type PspField = "taxPayerFee" | "pspBusinessName";
```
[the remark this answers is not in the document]
EmanueleBVtech: done here c2ec7b28eea99055f24bd4ffe767f580cd3bb719

src/utils/__tests__/sortPsp.test.ts
```diff
@@ -0,0 +1,57 @@
...
+test("sortBy function sorts bundles correctly in descending order", () => {
+  const sorted = JSON.parse(JSON.stringify(bundles)).sort(
+    sortBy("taxPayerFee", "desc")
+  ); // Deep copy
+  expect(sorted).toEqual([
+    { taxPayerFee: 200 },
+    { taxPayerFee: 150 },
+    { taxPayerFee: 100 },
+    { taxPayerFee: 0 }, // 0 should come before undefined
+    { taxPayerFee: undefined },
```
pietro-tota, line 40: are we sure that there are cases where fees are undefined? this should never happen since it will mean that for that specific bundle we don't know the fee that will be applied to the transaction 🤔

EmanueleBVtech: undefined is just a security measure just in case we have malformed data for any reason.

note that undefined will be always at the bottom of the list (both asc and desc) to avoid catch the user eye,
just in case

in general, better safe than sorry i guess.

src/__integration_tests__/utils/helpers.js
```diff
@@ -53,6 +53,40 @@ export const authorizePaymentAndGetError = async (
...
+*/
+export const checkPspList = async (noticeCode, fiscalCode, email, cardData) => {
+  const pspEditButtonSelector = "#pspEdit";
+  const pspFeeSortButtonId = "#sortByFee";
+
+  await fillAndSubmitCardDataForm(noticeCode, fiscalCode, email, cardData);
+
+  const pspEditButton = await page.waitForSelector(pspEditButtonSelector);
+  await pspEditButton.click();
+  await page.waitForTimeout(1000);
```
scaminati-bv, line 67: Don't use waitForTimeout because is [deprecated](https://github.com/cloudflare/puppeteer/blob/main/docs/api/puppeteer.page.waitfortimeout.md).
Btw don't use fixed timeout during test because you may be wasting time unnecessarily. 
For example waitForSelector wait only time is necessary until element appear.

EmanueleBVtech: this is not a mistake, you are missing one info. 😄 

we needed waitForTimeout with explicit time because we need to wait for a drawer menu animation to properly complete in order to execute this test correctly.

scaminati-bv: In this test we don't need animation, so you can disable it by adding this code on beforeAll test function:

```
page.on("load", () => {
    const content = `* {transition-duration: 0s !important;}`;
    page.addStyleTag({ content });
});
```

And then you could change test like this:

```
export const checkPspList = async (noticeCode, fiscalCode, email, cardData) => {
  const pspEditButtonSelector = "#pspEdit";
  const pspFeeSortButtonId = "#sortByFee";

  await fillAndSubmitCardDataForm(noticeCode, fiscalCode, email, cardData);
  const pspEditButton = await page.waitForSelector(pspEditButtonSelector);
  //open sidebar
  await pspEditButton.click();
  const pspFeeSortButton = await page.waitForSelector(pspFeeSortButtonId);
  await pspFeeSortButton.click();
  // get the list of divs
  const pspElements = await page.$$(".pspFeeValue");
  // Extract numeric content from each div and return as an array
  const numericContents = await Promise.all(
    Array.from(pspElements).map(async (element) => {
      const text = await element.evaluate((el) => el.textContent);
      // We want to skip the Dollar, Euro, or any currency placeholder
      let numbers = text.match(/[\d,]+/g); // This will match sequences of digits and commas
      let result = numbers ? numbers.join("").replace(",",".") : ""; // Join the matched numbers if any and replace , as separator with .
      return parseFloat(result) || 0; // Convert to number, default to 0 if NaN
    })
  );
  const closePspListButton = await page.waitForSelector("#closePspList");
  await closePspListButton.click();
  // wait for sidebar closure
  await page.waitForSelector(pspFeeSortButtonId, {hidden: true});
  return numericContents;
};
```

With this change you can speed up test execution

src/__integration_tests__/paymentflow.integration.test.ts
scaminati-bv, line 1: Please remove all formatter changes of old code (also on other files). It is difficult to check eventual mistake.

src/utils/__tests__/sortPsp.test.ts
```diff
@@ -0,0 +1,57 @@
...
+});
+
+test("sortBy function places undefined values at the end in ascending order", () => {
+  const sorted = JSON.parse(JSON.stringify(bundles)).sort(
+    sortBy("taxPayerFee", "asc")
+  ); // Deep copy
+  expect(sorted[sorted.length - 1].taxPayerFee).toBeUndefined();
+});
+
+test("sortBy function sorts bundles correctly in descending order", () => {
```
pietro-tota, line 31: add tests to check right sorting also for literal fields too (the case bundles are sorted by name

EmanueleBVtech: literal sorting is not affected by this PR, should we test it anyway?

pietro-tota: Why is not affected? Both sorting functionality are performed by the same method that have been touched here + better have an extra test than nothing

EmanueleBVtech: yes but the function only changed from this
`return fieldA && fieldB ? (fieldA > fieldB ? order : -order) : -order;`
to basically this
`fieldA !== undefined && fieldB !== undefined ? (fieldA > fieldB ? order : -order) : -order;`

this allows us to sort correctly value = 0 (the 0 was buggy since fieldA && fieldB fails when is 0 && 0) 

the literal sorting would never fail the fieldA && fieldB test since empty string for the name pspBusinessName can never be empty string. this is why i say it's not impacted (infact, the bug on the pspBusinessName never happened in the first place)

src/__integration_tests__/paymentflow.integration.test.ts
EmanueleBVtech: all tests involved in this PR are passed (is pretty much a guarantee there is no error there)
i will reduce the formatter footprint to allow quicker review :=)

src/__integration_tests__/utils/helpers.js
EmanueleBVtech: leaving the animation makes the test true to the real FE as-is. 
The test should not alter the behavior of the FE to become quicker or easier written.

I think we should not alter the behavior of the FE in any way.
(also the timeouts are a few seconds, wich i consider acceptable on a pipeline that anyway takes more than 15min to complete)

scaminati-bv: If you prefer to don't disable animations please remove deprecated method usage. 
image

Also i suppose there is too much timeout because you need only on open and on close action.
[Here](https://github.com/pagopa/pagopa-checkout-tests/blob/257761a09d8550c53d59af1fa1a353d42e794681/e2e-tests/src/npg/psp-sorting.npg.integration.test.ts) is used less timeout for similar test

src/__integration_tests__/paymentflow.integration.test.ts
EmanueleBVtech: done here 10097b7e6769916f969423bf89e6e4731142e093

src/__integration_tests__/utils/helpers.js
EmanueleBVtech: ok, if you insist i will replace with promise method, but i will not reduce timeout value since we did multiple tests and we feel comfortable with the current values.

again: we had an internal discussion and decided that few seconds of delay are not an issue

src/utils/__tests__/sortPsp.test.ts
pietro-tota: Yes I know, but theoretically speaking the code was touched and for a better coverage a test with string ordering will only add a guarantees about both ordering methods.
Moreover this will add a check for future development making test fail in case of any introduced regression. 
I don't see any problem adding an additional test, do you?

pietro-tota (approved):
LGTM

ciuffagianluca (approved):

infantesimone (approved):

PR #399, opened 2024-11-18
Feat [CHK-3353] Updated border on header

EmanueleBVtech:
#### List of Changes
Added border bottom to header on all pages to improve accessibility

#### Motivation and Context
We need to comply with accessibility requirements

Css edit does not require testing

#### Types of changes

#### Checklist:

### DIFF
--- a/src/components/commons/Header.tsx
+++ b/src/components/commons/Header.tsx
@@ -74,7 +74,11 @@ export default function Header() {
 
   return (
     <>
-      <Box p={3} bgcolor={"white"}>
+      <Box
+        p={3}
+        bgcolor={"white"}
+        sx={{ borderBottom: "1px solid", borderColor: "#EEEEEE" }}
+      >
         <Stack
           spacing={0}
           direction="row"


scaminati-bv (approved):

pietro-tota:
@ciuffagianluca ping :) can we merge it?

infantesimone (changes requested):
please check integration test

src/components/commons/SkipToContent.tsx
scaminati-bv, line 1: I can click the button with the mouse

EmanueleBVtech: well, this is as the pattern works
the link (button) becomes visible as soon as it gets focus
being clicked by the mouse provide focus and the button becomes visible

EmanueleBVtech: i made it harder to click here c4f395036d07acb90332c78a861292abe30fcd25

src/components/commons/SkipToContent.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <div style={{ position: "relative" }}>
+      <Button
+        href="#main_content"
+        onFocus={handleFocus}
+        onBlur={handleBlur}
+        style={linkStyle}
+      >
+        {t("mainPage.main.skipToContent")}
+      </Button>
```
scaminati-bv, line 28: If you click this button on riepilogo-pagamento page, this wrong dialog appear:

image

To fix this problem you should replace href attribute with `onClick={() => document.getElementById("main_content")?.focus()}`, because some pages have a check to prevent user leaving payment that is triggered by href.
This is appened beacuse href adds a fragment to the url, which triggers popstate event used to show this dialog.

src/components/commons/SkipToContent.tsx
```diff
@@ -0,0 +1,33 @@
...
+  };
+
+  return (
+    <div style={{ position: "relative" }}>
+      <Button
+        href="#main_content"
+        onFocus={handleFocus}
+        onBlur={handleBlur}
+        style={linkStyle}
+      >
```
scaminati-bv, line 26: Also add `tabIndex={-1}` in order to show the user where the focus has moved to after SkipToContent click.

src/components/commons/Header.tsx
```diff
@@ -90,6 +91,8 @@ export default function Header() {
             style={{ width: "56px", height: "36px" }}
             aria-hidden="true"
           />
+          <SkipToContent />
```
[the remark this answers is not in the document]
EmanueleBVtech: I don't think this is an issue.

I think having the button centered helps distinguish it from anything else in the toolbar, 
and allows for easier navigation for users with limited sight.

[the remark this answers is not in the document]
scaminati-bv: But the position of the button change based on page, may confuse users.

src/components/commons/SkipToContent.tsx
EmanueleBVtech: Not possible
As per documentation this would disable tab navigation twards the skip-to-content button too, wich is against the scope of the task
image

src/components/commons/SkipToContent.tsx
EmanueleBVtech: i'm afraid we cannot do that, because the screen reader needs the href to be set on the skip-to-content link to be able to bypass (eventual) text content. Remember from an html point of view, the skip to content needs an <a> tag with href set (at least according to the documentation found online).

src/components/commons/SkipToContent.tsx
scaminati-bv: It is correct that main content cannot be reached via tab navigation. This occurs only when you click the skipToContent button.

src/components/commons/SkipToContent.tsx
EmanueleBVtech: I'm thinking of a possible work around

src/components/commons/Header.tsx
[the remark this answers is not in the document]
EmanueleBVtech: ok i'll fix this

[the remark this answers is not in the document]
EmanueleBVtech: done here ecf0f7c0a5c73d1781fb0f3d955757ba2e66223e

src/components/commons/SkipToContent.tsx
EmanueleBVtech: done here e4eb02844ee93129becf1d979bc9a4e1dd6bfdee

src/components/commons/SkipToContent.tsx
EmanueleBVtech: done here e4eb02844ee93129becf1d979bc9a4e1dd6bfdee

.devops/pagopa-deploy-pipelines.yml
```diff
@@ -53,7 +53,7 @@ resources:
     - repository: checkoutTests
       type: github
       name: pagopa/pagopa-checkout-tests
-      ref: main
+      ref: CHK-3353-fix-integration-tests
```
scaminati-bv, line 56: Revert to main after [checkout-tests PR](https://github.com/pagopa/pagopa-checkout-tests/pull/103) merge

pietro-tota (approved):
LGMT, please resolve other reviewers comments

.devops/pagopa-deploy-pipelines.yml
EmanueleBVtech: done here f722b5fa0475bc6e022c7f0e6a227024268beedc

PR #401, opened 2024-11-18
Feat [CHK-3358] Added dynamic title

EmanueleBVtech:
#### List of Changes
Browser tab title changes according to page content

#### Motivation and Context
We need to comply with accessibility constraints

Manual test

image
image

also other langs
image

#### Types of changes

#### Checklist:


pietro-tota (approved):
LGTM
Please add some screenshot with some tabs with changed titles

scaminati-bv (approved):

src/components/PageContent/PageContainer.tsx
```diff
@@ -12,6 +12,12 @@ export default function PageContainer(props: {
 }) {
   const { t } = useTranslation();
 
+  useEffect(() => {
+    if (props.title) {
+      (document.title as any) = t(props.title) + " - pagoPA";
```
scaminati-bv, line 17: There are some pageContainer that do not have the title property. Check where it is absent and add it

EmanueleBVtech: the decision of using pageContainer titles as browser tab title has been taken with people from the design and UI team, we cannot make this decision by ourselfes. Current requisite is: when title inside the page changes, update the document title with the same text.

I need to check with them before doing this.

scaminati-bv: Ok check with design team, but consider that with this implementation if you change the page to other one without a title, it will show the previous one which may be wrong.
In any case (setting page title everywhere or not) It is better to choose a default title when it is not set in pageContainer

scaminati-bv: This is the problem that i reported before. This title is wrong on this page.
image

EmanueleBVtech: we checked with the design team and decided that pages with no title should not produce an update in the docuent title, you can check requirement here
https://pagopa.atlassian.net/browse/CHK-3358

EmanueleBVtech: I'm not sure. As per task requirement, pages that do not have a title should not update the document tab title.

EmanueleBVtech: done here 9233a9413700e6ce508437159981067ba9ab4dd4

scaminati-bv (approved):

PR #402, opened 2024-11-19
Bugfix [CHK-3566] solved sorting on PSP list using string values

EmanueleBVtech:
#### List of Changes
Minor edit to fix sorting on PSP using names

#### Types of changes

#### Checklist:


src/utils/SortUtil.ts
```diff
@@ -17,5 +17,11 @@ export const sortBy =
       return -1; // Place undefined at the end
     }
 
-    return fieldA > fieldB ? order : -order;
+    // Check if both values are strings
+    if (typeof fieldA === "string" && typeof fieldB === "string") {
+      return fieldA.localeCompare(fieldB) * order;
```
pietro-tota, line 22: please extends test cases for string comparison with lower/upper case mixed cases in order to have a test to check comparison, maybe reporting the same names for psps that originated the bug

EmanueleBVtech: done here c1eac42055f0c707986e064e4667bdfe2f41a815

scaminati-bv, line 22: By default **localCompare** does not guarantee case insensitive ordering.
According to [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator#sensitivity) use **base** option or if you prefear you can compare the two strings after toLowerCase()/toUpperCase() convertion

EmanueleBVtech: We should not use toLowerCase/toUpperCase for the following reason: localeCompare is the recommended approach when sorting JS arrays as it ensures that your sorting respects the nuances of different languages and character sets 
(takes into account the locale).

Added base fix tho.

scaminati-bv: I suggested using loweCase/upperCase with the localCompare function as-is.
`fieldA.toLowerCase().localeCompare(fieldB.toLowerCase())`.
BTW with parameter as you did is the best choice.

EmanueleBVtech: this is probably the absolute best
`fieldA.toLowerCase().localeCompare(fieldB.toLowerCase())`
but at this point i think we can do fine with the current code, right?

scaminati-bv: For me is equivalent, so use the version you prefear

EmanueleBVtech: if you are convinced, can you approve the PR?

scaminati-bv (approved):

ciuffagianluca (approved):

PR #408, opened 2024-12-09
Fix [CHK-3611] Fixed overflow scroll on payment drawer when long text happens

EmanueleBVtech:
#### List of Changes
Fixed overflow scroll on payment drawer when long text happens

#### Motivation and Context
Fixed overflow scroll on payment drawer when long text happens

New behavior
image

Previous behavior

image

#### Types of changes

#### Checklist:


ciuffagianluca (approved):

src/components/Header/DrawerCart.tsx
scaminati-bv, line 1: What do you think about setting this property in the upper div? In this way if the other fields are too long they are also handled.

src/components/Header/DrawerCart.tsx
```diff
@@ -111,7 +111,12 @@ export default function DrawerCart(props: Props) {
             >
               {t("cartDetail.description")}
             </Typography>
-            <Typography component="div" typography="sidenav" display="block">
+            <Typography
+              component="div"
+              typography="sidenav"
+              display="block"
+              style={{ overflowWrap: "break-word" }}
```
scaminati-bv, line 118: What do you think about setting this property in the upper div? In this way if the other fields are too long they are also handled.

EmanueleBVtech: The only fields that can cause this problem are {el.description} and {el.companyName}, since the other fields are fixed lenght fields (Notice code and Payee tax code), while the price cant ever reach the critical number of digits.

scaminati-bv: Ok perfect

scaminati-bv (approved):

PR #416, opened 2025-01-20
[CHK-3380] Added missing title to info modal + Collect other PRs into this to use 1 release

EmanueleBVtech:
This PR will collect also the following PR before merge:

- https://github.com/pagopa/pagopa-checkout-fe/pull/417
- https://github.com/pagopa/pagopa-checkout-fe/pull/418
- https://github.com/pagopa/pagopa-checkout-fe/pull/419
- https://github.com/pagopa/pagopa-checkout-fe/pull/420
- https://github.com/pagopa/pagopa-checkout-fe/pull/422
- https://github.com/pagopa/pagopa-checkout-fe/pull/423
- https://github.com/pagopa/pagopa-checkout-fe/pull/424
- https://github.com/pagopa/pagopa-checkout-fe/pull/425
- https://github.com/pagopa/pagopa-checkout-fe/pull/426

#### List of Changes
Added title to info modal to assist digital reader.
Added alt text for image

#### Motivation and Context
This is part of the accessibility changes

#### Screenshots (if appropriate):
image
#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #418, opened 2025-01-20
[CHK-3398] Added title to info modal

EmanueleBVtech:
#### List of Changes
The info modal in the page now has the title properly formatted

#### Motivation and Context
This is part of the accessibility update

Manual test

#### Types of changes

#### Checklist:


src/components/modals/InformationModal.tsx
```diff
@@ -36,7 +43,17 @@ function InformationModal(props: {
...
+        {props.title && (
+          <Typography
+            id={props.titleId}
+            variant="h6"
+            component={"div"}
+            sx={{ mb: 2 }}
+          >
+            {props.title}
+          </Typography>
+        )}
```
scaminati-bv, line 56: Instead of add a title and change current UI, what do you think about move close button outside dialogTitle? 
Like this [example](https://mui.com/material-ui/react-dialog/#customization) in the documentation page.

EmanueleBVtech: I dont understand your point. The issue here is not the close button. The problem is an empty h2 in the modal. I tested that adding a title does populate correctly the underlying h2 element.

Note: the link you provided does not point to an example.

scaminati-bv: By removing dialogTitle empty h2 element is not present anymore

EmanueleBVtech: Ok, i see. Yet, I think is just better to have the dialog ready to support the title as a feature. Having a title on the modal is a viable solution to the current issue and can be used later too if we ever need a title in the modal.

Both solutions are good i guess, i'd keep mine just because is ready and works.
What do you think about this?

scaminati-bv: The best solution is to use DialogTitle, but on other dialogs the title is present inside DialogContent. At the moment we can't do reafactor of all dialogs to align implementation (I think is out of scope of the task).
One more consideration is this fix works only for qrcode page, but empty h2 element remains in other dialogs. For this reason i prefear to remove DialogTitle in this component.

EmanueleBVtech: As you say, using the DialogTitle is the best solution. 
This is why i did it. 😉

You are missing info. All other dialogs (for wich we have an open task about the empty h2 in the title) have also already been corrected with same approach in their respective PR. 

This solution is solid and solves the issue. 
I don't see how the proposed solution adds value to the task.

I'm closing this for now.

scaminati-bv: If a title is present you should use DialogTitle, but in this case title is not expected by Design team. 
Adding title to this does not rispect figma. For this reason adding title to this modal is not the best choice.

Another consideration is that actual style of DialogTitle is different with the style used before.

EmanueleBVtech: Why you say the title is not already present?

The title is present here
image

and here
image

I just moved it in the right component.

scaminati-bv: I see the previous "Files Changed" that contain commit related to this [PR](https://github.com/pagopa/pagopa-checkout-fe/pull/416). I suppose they disappear after change base branch.

EmanueleBVtech: Its ok. Thanks for you deep feedback, is very usefull to hear your opinion.

pietro-tota (approved):

CianoDanilo (approved):

PR #438, opened 2025-02-18
[CHK-3715] feat: add login header until choose payment method page

scaminati-bv:
#### List of Changes
- add new top header with "login" button
- after clicking "login" a sample user is saved on session storage and on react state
- "login" button is visible until choose payment method page
- "logout" can be done until the end of the payment flow
- when payment flow ends, logged user is removed from session storage like other session data

#### Motivation and Context
The goal of this PR is to add header that contain login button with basic authentication flow. For this PR login action is mocked with hardcoded user sample.

Login
login

Logged user
logged

Logout
logout

#### Types of changes

#### Checklist:

### DIFF
--- a/src/components/commons/Header.tsx
+++ b/src/components/commons/Header.tsx
@@ -95,11 +95,15 @@ export default function Header() {
   };
 
   const initFeatureFlag = async () => {
-    await evaluateFeatureFlag(
-      featureFlags.enableAuthentication,
-      onFeatureFlagError,
-      onFeatureFlagSuccess
-    );
+    const storedFeatureFlag = getSessionItem(SessionItems.enableAuthentication);
+    // avoid asking again if you already have received an answer
+    if (!storedFeatureFlag) {
+      await evaluateFeatureFlag(
+        featureFlags.enableAuthentication,
+        onFeatureFlagError,
+        onFeatureFlagSuccess
+      );
+    }
   };
 
   useEffect(() => {
--- a/src/components/commons/Header.tsx
+++ b/src/components/commons/Header.tsx
@@ -117,7 +117,7 @@ export default function Header() {
         {enableAuthentication && <LoginHeader />}
         <Box p={3} bgcolor={"white"}>
           <Stack
-            spacing={0}
+            spacing={4}
             direction="row"
             justifyContent="space-between"
             alignItems="center"
--- a/src/components/commons/Header.tsx
+++ b/src/components/commons/Header.tsx
@@ -65,6 +65,9 @@ export default function Header() {
     CheckoutRoutes.ESITO,
     CheckoutRoutes.DONA,
   ];
+  const enablePaymentSummaryButton =
+    (!!PaymentInfo.receiver || !!CartInfo?.paymentNotices) &&
+    !ignoreRoutes.includes(currentPath);
   const toggleDrawer = (open: boolean) => {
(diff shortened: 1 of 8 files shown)


EmanueleBVtech:
can it make sense to add depends on https://github.com/pagopa/pagopa-checkout-fe/pull/439
as soon as that one is merged, you can integrate the **enableAuthentication** react state to display the button

scaminati-bv:
> can it make sense to add depends on #439 as soon as that one is merged, you can integrate the **enableAuthentication** react state to display the button

We can integrate the two PR after first merged because there's no one way dependency

src/components/commons/LoginHeader.tsx
```diff
@@ -0,0 +1,85 @@
...
+  );
+  const [loading, setLoading] = React.useState(false);
+  const onLoginClick = () => {
+    setLoading(true);
+    const user = {
+      id: "1234546",
+      email: "email@test.com",
+      name: "Mario",
+      surname: "Rossi",
+    };
```
pietro-tota, line 48: are this hardcoded info needed here?

scaminati-bv: This user is used to verify that the header displays the information correctly after login and will be removed. Would you prefer that this mock is removed immediately since we already check that header work correctly?

scaminati-bv (commented):
Set tag infra of **generate:checkout-feature-flags** script when https://github.com/pagopa/pagopa-infra/pull/2803 is merged

EmanueleBVtech (approved):

src/components/commons/Header.tsx
```diff
@@ -108,44 +112,43 @@ export default function Header() {
...
+            >
+              <img
+                src={pagopaLogo}
+                alt="pagoPA"
+                style={{ width: "56px", height: "36px" }}
+                aria-hidden="true"
+              />
+              <SkipToContent />
+            </Stack>
+            {(!!PaymentInfo.receiver || !!CartInfo?.paymentNotices) &&
```
sfarralorenzo, line 139: This logic is inherited from the previous code, but I suggest extracting it into a dedicated constant to simplify the rendering phase. For example:

```
const enablePaymentSummaryButton = 
  (!!PaymentInfo.receiver || !!CartInfo?.paymentNotices) &&
  !ignoreRoutes.includes(currentPath);
```

This way, the rendering code remains clean and focused:

```
{enablePaymentSummaryButton && (
  <Button
    onClick={() => toggleDrawer(true)}
    aria-label={t("mainPage.header.detail.detailButton")}
    endIcon={<ShoppingCart />}
  >
    {moneyFormat(amountToShow())}
  </Button>
)}
```

This approach also aligns with the code used just after:

```
{enableAuthentication && <Button>Mock login button</Button>}
```

sfarralorenzo (approved):
my comment is just a suggestion

src/components/commons/Header.tsx
scaminati-bv: Done here abd1d554990f54c944c24461df9f2bd9cdefe1ae

src/components/commons/LoginHeader.tsx
scaminati-bv: Removed here b8c322f39596f1ff74bf2240fe35beae6b86d6df

PR #441, opened 2025-02-20
[CHK-3714] Updated feature flag API signature

EmanueleBVtech:
#### List of Changes
Updated signature and usage of feature flag api to match new batch format

manually 
image

#### Types of changes

#### Checklist:

### DIFF
--- a/src/utils/api/helper.ts
+++ b/src/utils/api/helper.ts
@@ -1170,7 +1170,7 @@ export const evaluateFeatureFlag = async (
 ) => {
   await pipe(
     TE.tryCatch(
-      () => apiCheckoutFeatureFlags.evaluateFeatureFlag({ featureKey }),
+      () => apiCheckoutFeatureFlags.evaluateFeatureFlags({}),
       () => {
         onError("Network error");
         return new Error("Network error");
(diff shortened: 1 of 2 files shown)


dylantangredi-bvt (approved):

src/utils/api/helper.ts
```diff
@@ -1190,9 +1190,11 @@ export const evaluateFeatureFlag = async (
               return {};
             },
             (res: any) => {
-              if (res.value.enabled !== undefined) {
-                onResponse(res.value);
-                return res.value;
+              const target: any =
+                res.value !== undefined ? res.value[featureKey] : false;
```
sfarralorenzo, line 1194:
```suggestion
                res.value?.[featureKey] ?? false;
```

With the optional chaining we check both for `undefined` and `null` for both `res.value` and `res.value[featureKey]`. It preserves falsy but valid values like 0 and the empty string, which could be valid values for a feature flag.

Es.:
```
const dict = { 'mykey': 'myvalue' }
dict !== undefined ? dict['otherkey'] : false; // --> undefined`
// while
dict?.['otherkey'] ?? false; // --> false
```


If the idea was to also support `undefined` and `null` as valid values, then ignore the suggestion. Otherwise the suggested version seems more robust.

EmanueleBVtech: According to openfeature null & undefined are not common types supported by the specification
https://openfeature.dev/specification/types

I think we can ignore them, what do you think?

EmanueleBVtech: I think your option is better, so i will apply it.

sfarralorenzo (approved):

PR #442, opened 2025-02-20
[CHK-3717] feat: add authentication integration test

scaminati-bv:
#### List of Changes
- add authentication integration test

#### Motivation and Context
The goal of this PR is to verify that when the login button is clicked, the redirect url received from the login api is reached correctly.

Integration tests

#### Types of changes

#### Checklist:


.devops/azure-templates/setup-env.yml
```diff
@@ -52,6 +52,14 @@ parameters:
     type: string
     default: ''
 
+  - name: 'checkout_auth_service_host'
+    type: string
+    default: ''
+
+  - name: 'checkout_auth_service_baspath_v1'
```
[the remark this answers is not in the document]
scaminati-bv: These changes come from auth-service-api PR. After merge i check these changes again

.devops/azure-templates/setup-env.yml
```diff
@@ -108,6 +116,8 @@ steps:
       CHECKOUT_NPG_SDK_URL=${{ parameters.checkout_npg_sdk_url }} \
       CHECKOUT_GDI_CHECK_TIMEOUT=${{ parameters.checkout_gdi_check_timeout }} \
       CHECKOUT_VERSION=${{ parameters.checkout_version }} \
+      CHECKOUT_AUTH_SERVICE_HOST=${{ parameters.checkout_auth_service_host }} \
+      CHECKOUT_API_AUTH_SERVICE_BASEPATH_V1=${{ parameters.checkout_auth_service_baspath_v1 }} \
```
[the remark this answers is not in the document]
scaminati-bv: Same as above

dylantangredi-bvt (approved):

src/utils/api/client.ts
```diff
@@ -69,3 +69,12 @@ export const apiPaymentEcommerceClientWithRetryV2 = createEcommerceClientV2({
     async (r: Response): Promise<boolean> => r.status > 499
   ),
 });
+
+/**
+ * Api client for checkout auth service API V1
+ */
+export const apiCheckoutAuthServiceClientV1 = createAuthServiceClient({
```
sfarralorenzo, line 76: We need to add:
```
import { createClient as createAuthServiceClient } from "../../../generated/definitions/checkout-auth-service-v1/client";
```

because `createAuthServiceClient` is not imported

src/__integration_tests__/authentication.integration.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  it("Should correctly redirect to auth login url", async () => {
+
+    console.log("Search login button")
+    const loginHeader = await page.waitForSelector("#login-header");
+    const headerButtons = await loginHeader.$$("button");
+
+    //search login button and click it
+    Array.from(headerButtons).forEach(async (element) => {
+      const buttonText = await element.evaluate((el) => el.textContent);
+      if(buttonText == "Accedi") {
```
pietro-tota, line 34: can we check for button based on other condition rather than button text?

My concern here is to make test independent from translations

sfarralorenzo (changes requested):
great pr Simo.
I left a comment for a missing `import` statement

src/__integration_tests__/authentication.integration.test.ts
scaminati-bv: Unfortunately login button doesn't have id and i can't add it because is a component of `HeaderAccount `of mui-italia.
The only thing we could do is to take the last button present in the header, since it seems to me that there are no other buttons on the right after the login.
Do you think this approach might be better?

src/utils/api/client.ts
dylantangredi-bvt: I think that the import has been done in the auth-service-api PR, Simone correct me if I'm wrong.

scaminati-bv: This branch was created from auth-service-api PR so after merge a check again this one because there are more changes than I have made.

sfarralorenzo: Ok, I get that there are more PRs :)
Not trying to make a big deal out of this, absolutely. But we're reviewing _this specific PR_, so I have to comment that this code is not working if we don't add the import or we remove the client creation :) Do you agree?

scaminati-bv: Yes the bug is there so I will fix it in the main PR :)

scaminati-bv: Fixed here ec9911e92dce272f75d0ab6b155970c66637b2a8

scaminati-bv: fixed here ec9911e92dce272f75d0ab6b155970c66637b2a8

sfarralorenzo (approved):

src/__integration_tests__/authentication.integration.test.ts
scaminati-bv: Here 45191d672427915901b6a3077d347eea69a75782 i take the last button of the header in order to make test indipendent from translations.

ciuffagianluca (approved):

PR #445, opened 2025-02-24
[CHK-3734] add authentication callback route

scaminati-bv:
#### List of Changes
- add `auth-callback` page
- retrive auth-code from url query param
- save last page into session storage before login click
- go back to the saved page after retrieving the token

#### Motivation and Context
The goal of this PR is to add callback page where auth code is retrieving and then go back to saved page

#### Types of changes

#### Checklist:


src/routes/AuthCallbackPage.tsx
```diff
@@ -0,0 +1,49 @@
...
+      try {
+        window.addEventListener("beforeunload", onBrowserUnload);
+        window.addEventListener("popstate", onBrowserBackEvent);
+
+        //retrieve auth-code from url
+        const searchParams = new URLSearchParams(window.location.search);
+        const authCode = searchParams.get("auth-code");
+
+        //if auth code exists save into session storage
+        if(authCode != null)
```
infantesimone, line 22: What happens if `authCode` is null? In this case, invoking the `post auth` is unnecessary, so maybe we should display an error message and allow payment in guest mode?

scaminati-bv: If `authCode` is null error message popup is displayed without call api, but this implementation is added in the [PR](https://github.com/pagopa/pagopa-checkout-fe/pull/446) that extend this one with post auth api call.

infantesimone: OK, right, error handling is out of scope for the PR 💪

dylantangredi-bvt (approved):

infantesimone (approved):

PR #446, opened 2025-02-24
[CHK-3735] post auth implementation

ciuffagianluca:
Implement post authentication invocation and response OK management

#### Types of changes

#### Checklist:


src/utils/storage/sessionStorage.ts
```diff
@@ -72,6 +70,16 @@ export const getSessionItem = (item: SessionItems) => {
   }
 };
 
+export const getAndClearSessionItem = (item: SessionItems) => {
```
infantesimone, line 73:
```suggestion
export const getAndClearSessionItem = (item: SessionItems) => pipe(
  O.tryCatch(
    () => getSessionItem(item),  
    () => undefined 
  ),
  O.chain(value => {
    clearSessionItem(item); 
    return O.some(value);  
  })
);```

infantesimone (changes requested):

src/utils/storage/sessionStorage.ts
ciuffagianluca: done here ab4df85af4b2e34f19db199ac6db3150b631211d

EmanueleBVtech (approved):

PR #451, opened 2025-02-26
[CHK-3736] Handle post auth error

EmanueleBVtech:
#### List of Changes
The AuthCallback page now will display a full page message in case an error occours. 
Also, the user now can choose to try again the API call or go back to checkout without login.

#### Motivation and Context
This is part of a bigger effort to allow ID providers login on checkout-fe

Manual

image

#### Types of changes

#### Checklist:


src/routes/AuthCallbackPage.tsx
```diff
@@ -37,25 +41,34 @@ export default function AuthCallback() {
     );
   };
 
-  useEffect(() => {
-    try {
-      window.addEventListener("beforeunload", onBrowserUnload);
-      window.addEventListener("popstate", onBrowserBackEvent);
+  const doPostAuth = () => {
+    window.addEventListener("beforeunload", onBrowserUnload);
+    window.addEventListener("popstate", onBrowserBackEvent);
```
scaminati-bv, line 46: Move this `eventListener` on page init and set `removeEventListener` only before redirect

EmanueleBVtech: no those need to be in the function, because if **doPostAuth** fails, then the **onError** will remove those 2 listeners.
If the user clicks try again, we need those listener back again.

If they are inited on the useEffect and not at the beginning of **doPostAuth**, you might have the 2nd run of the function to go without those guards.

scaminati-bv: If you set `removeEventListener` only before redirect you don't need to set and remove on every api call

EmanueleBVtech: You are missing this part this error case:

If you only remove them before redirect, you have them ON as you land on the page and never remove them untill a user clicks a button to leave the page.

If the user decides to navigates to another page using browser arrows (or by going back in history) he will be blocked by the guard, even if no process is running.

dylantangredi-bvt (approved):

src/routes/AuthCallbackPage.tsx
EmanueleBVtech: I tried locally your suggestion, this is what happens in your scenario:
image

Is this a wanted behavior?
If this is not wanted, consider approving, thanks for the time. :)

scaminati-bv: I agree that if error is showing this alert is not necessary. So for me is ok your implementation.

EmanueleBVtech: Consider adding a check so we can merge into the other PR

scaminati-bv (approved):

PR #455, opened 2025-03-03
[CHK-3768] Retry button now repeats login

EmanueleBVtech:
#### List of Changes
 Retry button now repeats whole login

#### Motivation and Context
Refined requirements pointed in this direction

Manual test

#### Types of changes

#### Checklist:

### DIFF
--- a/src/routes/AuthCallbackPage.tsx
+++ b/src/routes/AuthCallbackPage.tsx
@@ -27,6 +27,8 @@ export default function AuthCallback() {
 
   const onError = () => {
     setLoading(false);
+    window.removeEventListener("popstate", onBrowserBackEvent);
+    window.removeEventListener("beforeunload", onBrowserUnload);
     ref.current?.reset();
   };
 
--- a/src/routes/AuthCallbackPage.tsx
+++ b/src/routes/AuthCallbackPage.tsx
@@ -1,10 +1,11 @@
-import React from "react";
+import React, { useEffect } from "react";
 import { useNavigate } from "react-router-dom";
 import { pipe } from "fp-ts/function";
 import * as O from "fp-ts/Option";
 import { Box, Button, CircularProgress, Typography } from "@mui/material";
 import { useTranslation } from "react-i18next";
 import ReCAPTCHA from "react-google-recaptcha";
+import { onBrowserBackEvent, onBrowserUnload } from "utils/eventListeners";
 import PageContainer from "../components/PageContent/PageContainer";
 import ko from "../assets/images/response-umbrella.svg";
 import {
@@ -12,14 +13,15 @@ import {
   getReCaptchaKey,
   getSessionItem,
   SessionItems,
+  setSessionItem,
 } from "../utils/storage/sessionStorage";
-import { proceedToLogin } from "./../utils/api/helper";
+import { authentication, proceedToLogin } from "./../utils/api/helper";
 import { CheckoutRoutes } from "./models/routeModel";
 
 export default function AuthCallback() {
   const navigate = useNavigate();
   const ref = React.useRef<ReCAPTCHA>(null);
-  const [loading, setLoading] = React.useState(false);
+  const [loading, setLoading] = React.useState(true);
 
   const { t } = useTranslation();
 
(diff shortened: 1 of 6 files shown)


scaminati-bv (changes requested):
Why has the **doPostAuth** function and related methods for retrieving the authentication token been removed?

ciuffagianluca (dismissed):

ciuffagianluca (approved):

src/routes/AuthCallbackPage.tsx
```diff
@@ -4,30 +4,55 @@ import { pipe } from "fp-ts/function";
...
   const [loading, setLoading] = React.useState(true);
-  const [errorOnPostAuth, setErrorOnPostAuth] = React.useState(false);
 
   const { t } = useTranslation();
 
   const onError = () => {
     setLoading(false);
-    window.removeEventListener("popstate", onBrowserBackEvent);
-    window.removeEventListener("beforeunload", onBrowserUnload);
-    setErrorOnPostAuth(true);
```
scaminati-bv, line 30: Without this the error message will never be displayed

EmanueleBVtech: not really. 
the page is either loading or displayng the message, no need for another state.
Also, the event listener must be removed you are right.

src/routes/AuthCallbackPage.tsx
```diff
@@ -4,30 +4,55 @@ import { pipe } from "fp-ts/function";
...
+        window.location.assign(url);
+      }
+    } catch {
+      onError();
+    }
+  };
+
+  const onLogin = async (recaptchaRef: ReCAPTCHA) => {
+    setLoading(true);
+    await proceedToLogin({ recaptchaRef, onError, onResponse });
```
scaminati-bv, line 51: Using the same callbacks as the Post auth token could create problems because the response is different from the login api.

EmanueleBVtech: onError takes no input so there cannot be any conflict, 
and onResponse  is different from both. 

What is the issue more precisely, i don't see it

scaminati-bv (approved):

infantesimone (changes requested):
can you add unit test related to AuthCallbackPage.tsx?

EmanueleBVtech:
> can you add unit test related to AuthCallbackPage.tsx?

Unit tests are already present, this PR does not add content, it just changed the callback of the "Retry button".
Unit tests were already added in the previous PR

infantesimone:
> > can you add unit test related to AuthCallbackPage.tsx?
> 
> Unit tests are already present, this PR does not add content, it just changed the callback of the "Retry button". Unit tests were already added in the previous PR

There are no unit tests for this file. The issue is that it's a technical debt from previous PRs. It's time to stop adding technical debt. :-)

EmanueleBVtech:
> AuthCallbackPage



> > > can you add unit test related to AuthCallbackPage.tsx?
> > 
> > 
> > Unit tests are already present, this PR does not add content, it just changed the callback of the "Retry button". Unit tests were already added in the previous PR
> 
> There are no unit tests for this file. The issue is that it's a technical debt from previous PRs. It's time to stop adding technical debt. :-)

Ok, but, we never did unit tests for entire pages, i think integration tests are needed in this case.
 
 I mean, aren't unit tests more to test logic of some functions?
 
 Can i ask what type of unit test you were thinking of?

infantesimone:
> > AuthCallbackPage
> 
> > > > can you add unit test related to AuthCallbackPage.tsx?
> > > 
> > > 
> > > Unit tests are already present, this PR does not add content, it just changed the callback of the "Retry button". Unit tests were already added in the previous PR
> > 
> > 
> > There are no unit tests for this file. The issue is that it's a technical debt from previous PRs. It's time to stop adding technical debt. :-)
> 
> Ok, but, we never did unit tests for entire pages, i think integration tests are needed in this case.
> 
> I mean, aren't unit tests more to test logic of some functions?
> 
> Can i ask what type of unit test you were thinking of?

Added unit test https://github.com/pagopa/pagopa-checkout-fe/pull/455/commits/c54da9c691751c841f92fb09b7ab61cb19c83cda

infantesimone (approved):

PR #456, opened 2025-03-05
[CHK-3763] feat: retrive user info from auth token

scaminati-bv:
Depends on [pagopa/pagopa-infra#2864](https://github.com/pagopa/pagopa-infra/pull/2864)
Depends on [pagopa/pagopa-checkout-be-mock#96](https://github.com/pagopa/pagopa-checkout-be-mock/pull/96)

#### List of Changes
- add api call to `GET /auth/users` into `LoginHeader` component and `AuthCallback` page
- add retry for  `GET /auth/users` for all error statuses
- if getIUserInfo fails in `AuthCallback` an error message is displayed with the login failed message
- if getIUserInfo fails in `LoginHeader` after page refresh, user is redirected into `errorPage` and the authentication session is lost
- fix: `Continua senza login.` label by removing final dot
- fix: After login on initial page feature flag enableAuthentication is lost
- fix: On `AuthCallback` page, after login retry, error message was displayed for a moment

#### Motivation and Context
The goal of this PR is to retrieve user info from `GET /auth/users` api call. This api return name and surname of user related to authToken stored on SessionStorage and these values is saved into redux state in order to share this informations cross components (AuthCallback page and LoginHeader component).

Unit test

#### Types of changes

#### Checklist:


ciuffagianluca (approved):

src/components/commons/LoginHeader.tsx
```diff
@@ -39,20 +49,25 @@ export default function LoginHeader() {
...
+  const [isLoginButtonReady, setLoginButtonReady] = React.useState(false);
   const [error, setError] = React.useState("");
   const [errorModalOpen, setErrorModalOpen] = React.useState(false);
 
-  const onError = (m: string) => {
+  // the login button should be visible if user is already logged in
+  // and user is on pages where he cannot do login
+  const showLoginButton = (): boolean =>
+    isLoginButtonReady &&
+    (loginRoutes.includes(currentPath) || loggedUser.userInfo != null);
```
EmanueleBVtech, line 63: This logic `loggedUser.userInfo != null` testing the presence of a userInfo object will also be true also when userInfo is undefined (which is the initial redux state). 

Was this intended? 
If not i can suggest something like `!loggedUser.userInfo` to validate against both null and undefined

scaminati-bv: [Loose Equality !=](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#loose_equality_using)  works with both null and undefined.

> If one of the operands is null or undefined, the other must also be null or undefined to return true. Otherwise return false.

src/components/commons/__tests__/LoginHeader.test.tsx
```diff
@@ -0,0 +1,139 @@
...
+      <MemoryRouter>
+        <LoginHeader />
+      </MemoryRouter>
+    );
+    await waitFor(() => {
+      const loginButton = screen.getByTitle(/Accedi/i);
+      expect(loginButton).toBeInTheDocument();
+      fireEvent.click(loginButton);
+      expect(proceedToLogin).toHaveBeenCalled();
+      expect(location.href).toBe(redirectUrl);
```
EmanueleBVtech, line 93: In the second test, location.href is directly checked for the redirection. Since Jest test environments may not trigger actual redirects or modify the global location.href automatically, this could cause a failure if the test setup doesn't account for it. Consider mocking the location object as follows:

`jest.spyOn(global, 'location', 'get').mockImplementation(() => ({
  href: ''
}));`

scaminati-bv: This plugin is used to mock all location object [jest-location-mock](https://www.npmjs.com/package/jest-location-mock).

EmanueleBVtech:
Looks good to me, 
maybe just consider adding a helper function to avoid duplication of code:

`
const renderLoginHeader = () =>
  renderWithReduxProvider(
    <MemoryRouter>
      <LoginHeader />
    </MemoryRouter>
  );
`

EmanueleBVtech:
to fix this you may want to use something like this

```
    const BASE_CALLBACK_URL_REGEX = "http:\\/\\/localhost:\\d+\\/auth-callback\\?code=([a-zA-Z0-9]+)&state=([a-zA-Z0-9]+)";
    const regex = new RegExp(BASE_CALLBACK_URL_REGEX);
    expect(regex.test(currentUrl)).toBe(true);
```

ciuffagianluca (dismissed):

CianoDanilo (approved):

package.json
```diff
@@ -6,7 +6,7 @@
...
-    "generate:checkout-auth-service-v1": "rimraf generated/definitions/checkout-auth-service-v1 && shx mkdir -p generated/definitions/checkout-auth-service-v1 && gen-api-models --api-spec https://raw.githubusercontent.com/pagopa/pagopa-infra/v1.404.1/src/domains/checkout-app/api/checkout/checkout_auth_service/v1/_openapi.json.tpl --no-strict --out-dir ./generated/definitions/checkout-auth-service-v1 --request-types --response-decoders --client",
+    "generate:checkout-auth-service-v1": "rimraf generated/definitions/checkout-auth-service-v1 && shx mkdir -p generated/definitions/checkout-auth-service-v1 && gen-api-models --api-spec https://raw.githubusercontent.com/pagopa/pagopa-infra/c0d12ccd9e29a7b67758288edb50e538df81cae4/src/domains/checkout-app/api/checkout/checkout_auth_service/v1/_openapi.json.tpl --no-strict --out-dir ./generated/definitions/checkout-auth-service-v1 --request-types --response-decoders --client",
```
scaminati-bv, line 9: Set tag after [PR](https://github.com/pagopa/pagopa-infra/pull/2864) is merged

scaminati-bv: Done here 3b5b7b552b0478e9249f994aad111947d3a7d044

ciuffagianluca (approved):

src/utils/api/client.ts
```diff
@@ -82,15 +82,19 @@ export const apiCheckoutAuthServiceClientV1 = createAuthServiceClient({
...
   fetchApi: retryingFetch(
     fetch,
     conf.CHECKOUT_API_TIMEOUT as Millisecond,
     3,
-    (response: Response) => response.status < 200 && response.status >= 400
+    (response: Response) =>
+      response.status < 200 &&
+      response.status >= 400 &&
+      response.status !== 401 &&
+      response.status !== 429
```
CianoDanilo, line 98: Why not only in case of 5xx?

scaminati-bv: I initially thought of all the errors and then excluded some of them. but it actually makes more sense to do retries only on 5xx. Done here 8e87a80999c871450d77870f8a408cf025682ed1

CianoDanilo (approved):

ciuffagianluca (approved):

CianoDanilo (approved):

ciuffagianluca (approved):

PR #470, opened 2025-03-17
[CHK-3806] fix: get user info after logout 

scaminati-bv:
#### List of Changes
- if `authToken `on `sessionStorage `is not present, discard get user response of LoginHeader (both success and error case)
- change `window.location.replace` with `navigate` in order to prevent app to be refreshed on redirect to internal page

#### Motivation and Context
The goal of this PR is to fix a racing condition of **get users** with **logout** after a payment error.
On error page a get user is performed despite having previously logged out.

On error page logged user must not be present

Before 
localhost_1234_errore

After
localhost_1234_errore (1)

#### Types of changes

#### Checklist:

### DIFF
--- a/src/components/commons/LoginHeader.tsx
+++ b/src/components/commons/LoginHeader.tsx
@@ -174,14 +174,23 @@ export default function LoginHeader() {
   const doGetUserInfo = () => {
     void retrieveUserInfo({
       onResponse: (userInfo: UserInfoResponse) => {
-        dispatch(
-          setLoggedUser({
-            id: userInfo.userId,
-            name: userInfo.name,
-            surname: userInfo.familyName,
-          })
+        pipe(
+          getSessionItem(SessionItems.authToken),
+          O.fromNullable,
+          O.fold(
+            () => setLoginButtonReady(true),
+            () => {
+              dispatch(
+                setLoggedUser({
+                  id: userInfo.userId,
+                  name: userInfo.name,
+                  surname: userInfo.familyName,
+                })
+              );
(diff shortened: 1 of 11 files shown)


src/components/commons/LoginHeader.tsx
```diff
@@ -174,14 +174,23 @@ export default function LoginHeader() {
   const doGetUserInfo = () => {
     void retrieveUserInfo({
       onResponse: (userInfo: UserInfoResponse) => {
-        dispatch(
-          setLoggedUser({
-            id: userInfo.userId,
-            name: userInfo.name,
-            surname: userInfo.familyName,
-          })
+        pipe(
```
pietro-tota, line 177: here you're discarding UserInfoResponse based on the session token presence into session storage that is ok but how this resolve the _racing condition_ where a 401 is received because in the "esito" page the logout is performed concurrently with get user info api call?
An error (not only the 401, any error) in the get user info cannot redirect to error page if we hare in the final result page because this will overwrite the outcome displayed to the user so I think that here should be at least manipulated the onError section of this section.
what do you think?

scaminati-bv: I agree with you, this check must be done also onError case.
Done here 6e718302896d4c6cfc80f7ff2b12ca827416d918

pietro-tota (commented):
LGMT 🚀

infantesimone:
please check the integration tests

infantesimone (approved):

pietro-tota (approved):

PR #475, opened 2025-03-25
[CHK-3867] Updated payment-flow navigation to include psp picker page based on ff

EmanueleBVtech:
#### List of Changes
Updated payment-flow navigation to include psp picker page based on ff
Inludes https://github.com/pagopa/pagopa-checkout-fe/pull/476

#### Motivation and Context
This is part of an improvements for the user experience

integration tests included in different pr

#### Types of changes

#### Checklist:


src/App.tsx
```diff
@@ -76,7 +78,35 @@ export function App() {
...
+      "Error while getting feature flag " + SessionItems.enablePspPage,
+      e
+    );
+  };
+
+  const onFeatureFlagSuccess = (data: { enabled: boolean }) => {
+    // we need to use localstorage to be permanent in case of page refreshes
+    // which happen after entering the credit card data
+    localStorage.setItem(SessionItems.enablePspPage, data.enabled.toString());
+  };
```
scaminati-bv, line 93: You can use sessionStorage by adding exception on `clearStorageAndMaintainAuthData` method on [SessionStorage](https://github.com/pagopa/pagopa-checkout-fe/blob/main/src/utils/storage/sessionStorage.ts) file.
Session storage is cleared only on first page and outcome/error pages.

src/App.tsx
```diff
@@ -76,7 +78,35 @@ export function App() {
...
+
+  const initFeatureFlag = async () => {
+    const storedFeatureFlag = localStorage.getItem(SessionItems.enablePspPage);
+    // avoid asking again if you already have received an answer
+    if (!storedFeatureFlag) {
```
scaminati-bv, line 98: With this condition feature flag is not evaluated anymore. Use sessionStorage instead or remove this condition to make api call on every page load.

EmanueleBVtech: session storage cannot be used due reasons discussed in other comment.
i added a logic to clear the enablePspPage when the sessionStorage is resetted

src/App.tsx
```diff
@@ -76,7 +78,32 @@ export function App() {
...
+      "Error while getting feature flag " + SessionItems.enablePspPage,
+      e
+    );
+  };
+
+  const onFeatureFlagSuccess = (data: { enabled: boolean }) => {
+    // we need to use localstorage to be permanent in case of page refreshes
+    // which happen after entering the credit card data
+    localStorage.setItem(SessionItems.enablePspPage, data.enabled.toString());
+  };
```
scaminati-bv, line 93: How about adding the authentication feature flag check here as well? Because for now the call is repeated twice.

src/features/payment/components/PaymentChoice/PaymentChoice.tsx
```diff
@@ -78,7 +78,12 @@ export function PaymentChoice(props: {
     }
 
     setLoading(false);
-    navigate(`/${route || CheckoutRoutes.RIEPILOGO_PAGAMENTO}`);
+
+    const correctedRoute =
+      localStorage.getItem(SessionItems.enablePspPage) === "true"
+        ? CheckoutRoutes.LISTA_PSP
+        : CheckoutRoutes.RIEPILOGO_PAGAMENTO;
+    navigate(`/${route || correctedRoute}`);
```
scaminati-bv, line 86: This navigate always to `route` variable and ignore feature flag.

src/App.tsx
scaminati-bv: Yes but we can also consider making this improvement in another PR

src/features/payment/components/PaymentChoice/PaymentChoice.tsx
EmanueleBVtech: good catch,

src/App.tsx
EmanueleBVtech: ok, will be discussed in different pr then, closing this one

src/features/payment/components/PaymentChoice/PaymentChoice.tsx
```diff
@@ -78,7 +78,17 @@ export function PaymentChoice(props: {
     }
 
     setLoading(false);
-    navigate(`/${route || CheckoutRoutes.RIEPILOGO_PAGAMENTO}`);
+
+    const correctedRoute = route || CheckoutRoutes.RIEPILOGO_PAGAMENTO;
```
scaminati-bv, line 82: Instead of instantiate another variable use directly `route` by adding fallback here:

`const route: string = PaymentMethodRoutes[paymentTypeCode]?.route || CheckoutRoutes.RIEPILOGO_PAGAMENTO;`

scaminati-bv (approved):

src/features/payment/components/PaymentChoice/PaymentChoice.tsx
EmanueleBVtech: what does this achieve?

scaminati-bv: Code readability and maintainability by removing useless variable

EmanueleBVtech: having the route and correctedRoute variables separated actually gives visibility of the actual value of the route variable during debugging and development, if we remove the variable, it becomes impossible to know if we did use a fallback on the RIEPILOGO_PAGAMENTO, so i dont think is useless.

Do you aggree?

scaminati-bv: For me it is not important to know if fallback is used, so I would simplify the code. But if you prefear leave it like that.

EmanueleBVtech: ok lets see what other think
:)

scaminati-bv (commented):
By now psp-list page is accessible from url also if featureFlag is false. It would be better to add check in the guard.

EmanueleBVtech:
> By now psp-list page is accessible from url also if featureFlag is false. It would be better to add check in the guard.

i think this was done by design in the other pr to allow debugging of the page without having access to the page via payment flow. Realistically, only we now that page exists, people would need to guess the url?

src/features/payment/components/PaymentChoice/PaymentChoice.tsx
pietro-tota: I would leave those variables separate but maybe refactor from `correctRoute` to something like `navigateToRoute`, just because it seems that we don't follow the correct route if the ff is enabled :D

EmanueleBVtech: done

pietro-tota (dismissed):

sfarralorenzo (approved):

scaminati-bv (approved):

pietro-tota (approved):

CianoDanilo (approved):

PR #489, opened 2025-04-03
[CHK-2498] feat: Upgrate to react 18

EmanueleBVtech:
#### List of Changes
Upgraded the following: 
- @mui/material to 5.16.14 
- react-router-dom to 6.29.0 
- react to 18.3.1, i18next to 22.5.1
- react-google-recaptcha to 3.1.0
- updated types declarations
- updated integration tests

Edits:
- session storage clear use delayed scheduled execution to work properly in the react 18 component update events tree
- Fixed some type issues here and there (react uses stricter type checks)

Added the following modules to bridge type compatibility with node libraries:
- src/react-google-recaptcha.d.ts
- src/react-qr-reader.d.ts

react-google-recaptcha has a well know issue with react18 discussed [here](https://github.com/dozoisch/react-google-recaptcha/issues/250), and the proposed stable solution is to remove the strict mode from react. Another potential solution is to not use (yet) the concurrent rendering feature of react 18, to avoid a broken recaptcha.

#### Motivation and Context
We need to upgrade react version from 17 to 18

Manual testing on dev
image

#### Types of changes

#### Checklist:


src/__integration_tests__/authentication.integration.test.ts
```diff
@@ -656,9 +655,12 @@ describe("Logout tests", () => {
...
     await userButton.click();
-    const logoutButton = await page.waitForXPath('/html/body/div[5]/div[3]/ul/li');
-    console.log("wait for logout button");
-    await logoutButton.click();
+    await page.waitForSelector("#logout-button-icon");
+    await page.evaluate(()=>{
+      setTimeout(()=>{
+        document.getElementById("logout-button-icon").parentNode.click();
+      },100);
+    })
```
scaminati-bv, line 663:
```suggestion
    const logoutButtonIcon = await page.waitForSelector("#logout-button-icon");
    const logoutButton = await logoutButtonIcon.getProperty('parentNode');
    await logoutButton.click();
```

What do you think about this rather than adding evaluate with setTimeout?

scaminati-bv: Similar implementation is used into "Should invoke auth-service api with x-rpt-id header" test.

EmanueleBVtech: ok, done

src/utils/storage/sessionStorage.ts
```diff
@@ -114,7 +114,12 @@ export function clearSessionItem(name: SessionItems) {
 export const isStateEmpty = (item: SessionItems) => !getSessionItem(item);
 
 export const clearStorage = () => {
-  sessionStorage.clear();
+  // with react 18 the state update in components follows
+  // a different refresh behavior, and we need this delay to make sure
+  // this function happens as last action
+  setTimeout(() => {
+    sessionStorage.clear();
+  }, 100);
```
scaminati-bv, line 122: Instead of adding timeout and make the behavior unpredictable, convert const variables into react state objects, in order to mantains values on component re-rendering.
For example  in tu `PaymentResponsePage` do the following change:

From
`
const usefulPrintData: PrintData = {
    useremail: email || "",
    amount: moneyFormat(totalAmount),
  };
`
To
`
const [usefulPrintData] = useState<PrintData>({
    useremail: email || "",
    amount: moneyFormat(totalAmount),
  });
` 

Make these changes in all pages where the storage is cleared.

EmanueleBVtech: good idea, will do.
just to clarify tho, i don't see how the setTimeout makes this unpredictable, i've tested that setTiemout with times as low as 10 ms will still do the trick. 100 is a safe treshold.

changing it tho, i prefer the states as you suggest

EmanueleBVtech: done

scaminati-bv: The problem is not the value of timeout (I try 0 and it works in this case) but that this workaround may not work in future usage of this method.

EmanueleBVtech: The trick is that JS is single threaded, and the setTimeout appears to be triggered always after the react state update flow, so this is just a low-level tweak to force JS to run after "react is done doing its think". Should be future proof, but again, your idea is plain better.

src/react-google-recaptcha.d.ts
```diff
@@ -0,0 +1,25 @@
+declare module "react-google-recaptcha" {
```
pietro-tota, line 1: why are those needed?

infantesimone (approved):

pietro-tota (approved):

scaminati-bv (approved):

ciuffagianluca (approved):

PR #503, opened 2025-04-11
[TEST] fix: remove timeout from integration tests

scaminati-bv:
#### Types of changes

#### Checklist:


src/__integration_tests__/paymentflow.integration.test.ts
```diff
@@ -84,18 +84,6 @@ beforeAll(async () => {
 
 beforeEach(async () => {
   await page.goto(CHECKOUT_URL);
-  // Listen for dialog events and automatically accept them after a delay
-  page.on('dialog', async dialog => {
```
EmanueleBVtech, line 88: instead of removing this part, maybe just comment it out,
because errors will appear again after react concurrent rendering is enabled again

scaminati-bv: If this dialog during normal payment flow is not appears, for me is not correct to manage it on this test. 
The test should be manage in different way to avoid this wrong scenario when concurrent rendering is enabled.

EmanueleBVtech: I think its fair to leave that (functioning) code as potential solution for future reference and
let the decision to who ever will come next. 
:)

pietro-tota: imho there is no need to leave this trick here that auto accept dialog behind the hood: if a dialog appear during test it should be handled by the test itself

pietro-tota (approved):

sfarralorenzo (approved):

PR #511, opened 2025-05-13
fix: [PIDM-4123] removed mixpanel events checkout

FrancescoDiTonnoJakala:
#### Types of changes

#### Checklist:

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Editor-based HTTP Client requests
+/httpRequests/
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
--- a/src/components/commons/SurveyLink.tsx
+++ b/src/components/commons/SurveyLink.tsx
@@ -3,19 +3,12 @@ import { Typography, Alert, AlertTitle, Link, Box } from "@mui/material";
 import { useTranslation } from "react-i18next";
 import { ThemeProvider } from "@mui/material";
 import { theme } from "@pagopa/mui-italia";
-import { VOC_USER_EXIT } from "../../utils/config/mixpanelDefs";
-import { mixpanel } from "../../utils/config/mixpanelHelperInit";
 
 const SurveyLink = () => {
   const { t } = useTranslation();
 
   const onClick = () => {
-    mixpanel.track(VOC_USER_EXIT.value, {
-      EVENT_ID: VOC_USER_EXIT.value,
-      event_category: "UX",
-      event_type: "exit",
-      screen_name: "payment response page",
-    });
+    // ??
   };
 
   return (
--- a/src/routes/DonationPage.tsx
+++ b/src/routes/DonationPage.tsx
@@ -36,11 +36,6 @@ import {
 } from "../features/payment/models/donationModel";
 import { useAppDispatch } from "../redux/hooks/hooks";
 import { getDonationEntityList } from "../utils/api/helper";
-import {
-  DONATION_APPIO_VISIT,
-  DONATION_URL_VISIT,
-} from "../utils/config/mixpanelDefs";
-import { mixpanel } from "../utils/config/mixpanelHelperInit";
 import { moneyFormat } from "../utils/form/formatters";
 
 export default function DonationPage() {
@@ -85,9 +80,6 @@ export default function DonationPage() {
   };
 
   const onAPPIO = React.useCallback(() => {
-    mixpanel.track(DONATION_APPIO_VISIT.value, {
-      EVENT_ID: DONATION_APPIO_VISIT.value,
-    });
     if (isMobileDevice) {
       window.open("https://io.italia.it/donazioni-ucraina/", "_blank")?.focus();
     } else {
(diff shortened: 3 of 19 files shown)


sfarralorenzo:
I think we should add `.idea` in `.gitignore`, right?

.idea/.gitignore
pietro-tota, line 1: please delete those .idea files

pietro-tota (changes requested):

src/components/commons/SurveyLink.tsx
```diff
@@ -30,16 +19,6 @@ const SurveyLink = () => {
           <Typography variant="body1" mb={2}>
             {t("paymentResponsePage.survey.body")}
           </Typography>
-          <Link
```
pietro-tota, line 33: why are you removing this?

pietro-tota (changes requested):

src/components/commons/SurveyLink.tsx
FrancescoDiTonnoJakala: Solved by https://github.com/pagopa/pagopa-checkout-fe/pull/511/commits/8b5bed54271270c656915ff37bfee3edd7ce69a6

src/components/commons/SurveyLink.tsx
```diff
@@ -3,19 +3,12 @@ import { Typography, Alert, AlertTitle, Link, Box } from "@mui/material";
...
   const { t } = useTranslation();
 
   const onClick = () => {
-    mixpanel.track(VOC_USER_EXIT.value, {
-      EVENT_ID: VOC_USER_EXIT.value,
-      event_category: "UX",
-      event_type: "exit",
-      screen_name: "payment response page",
-    });
+    alert("click");
```
infantesimone, line 11: ?

infantesimone (changes requested):

src/components/commons/SurveyLink.tsx
FrancescoDiTonnoJakala: Solved by https://github.com/pagopa/pagopa-checkout-fe/pull/511/commits/36fe956bbdcb470ba68d980634b06fb00f6ee25d

ciuffagianluca (dismissed):

src/components/commons/SurveyLink.tsx
```diff
@@ -3,19 +3,13 @@ import { Typography, Alert, AlertTitle, Link, Box } from "@mui/material";
...
 
   const onClick = () => {
-    mixpanel.track(VOC_USER_EXIT.value, {
-      EVENT_ID: VOC_USER_EXIT.value,
-      event_category: "UX",
-      event_type: "exit",
-      screen_name: "payment response page",
-    });
+    // eslint-disable-next-line no-console
+    console.log("click");
```
infantesimone, line 12: @FrancescoDiTonnoJakala console.log in production with 'click', better to avoid it

sfarralorenzo (changes requested):
An event handling change could not be wanted

src/routes/DonationPage.tsx
```diff
@@ -134,13 +126,6 @@ export default function DonationPage() {
           rel="noopener noreferrer"
           sx={{ display: "flex", mx: 1 }}
           title={`${t("ariaLabels.informationLink")} ${entity.companyName}`}
-          onClick={(e) => {
-            e.stopPropagation();
```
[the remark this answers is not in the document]
ciuffagianluca: Here, all the on click handler is going to be removed. This makes sense if you note that the only event listener task is to add a mixpanel tracking event. So, by my side, it is right to remove all onClick listener

src/components/commons/SurveyLink.tsx
ciuffagianluca: Maybe is better to use an empty arrow function and disable a possible lint warning?

infantesimone: Let's remember fp-ts :-) 

```
import { constVoid } from "fp-ts/function";
const onClick = constVoid;
```

src/routes/DonationPage.tsx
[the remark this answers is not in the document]
sfarralorenzo: Sorry Gianlu I still I think we have to keep the `stopPropagation`. Without explicitly blocking event propagation, clicking on the Link/InfoIcon will cause the event to bubble up to the parent `FieldContainer`, triggering its `onClick` handler, right?

In the original implementation, `e.stopPropagation()` was explicitly included to prevent this bubbling, ensuring that clicking the info icon only opens the website without also selecting the entity. Removing this line could create a side effect where users would unintentionally select an entity while simply trying to view more information. This is why I wrote:

> If we are not completely sure that this change does not have side-effects, I would re-introduce the e.stopPropagation()

That said, maybe the page is not used anymore?
:)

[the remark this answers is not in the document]
FrancescoDiTonnoJakala: Solved by https://github.com/pagopa/pagopa-checkout-fe/pull/511/commits/e4c89db0fd2387f9e905591a46458890aa1fd743

sfarralorenzo (approved):

infantesimone (approved):

ciuffagianluca (approved):

ciuffagianluca:
By pass rules since mock updates make integration tests and code review pipeline systematically fails

PR #514, opened 2025-05-14
fix:[PIDM-544]replace <meta name="robots" content="noindex, nofollow" /> with <meta…

forciniti-luigi-jakala:
… name="robots" content="index, follow">

#### List of Changes

Replace 'meta name="robots" content="noindex, nofollow"'  with 'meta name="robots" content="index, follow"'

Added a step in the deployment pipeline for the dev and uat environments to prevent indexing by search engines.

#### Motivation and Context

To be done only in production

#### Types of changes

#### Checklist:


sfarralorenzo:
if this change should be just for the production environment, we should find a dynamic approach based on the environment, right?
Javascript is not an option because search engines could not executed it or execute it with a delay.
One approach could be adding a step in the dev/uat pipelines that modifies the index.html (production has to stay safe)?

So, before the build script, having something like:
```
# Change from production default "index, follow" to "noindex, nofollow"
    sed -i '' 's/<meta name="robots" content="index, follow">/<meta name="robots" content="noindex, nofollow">/g' ./src/index.html
```

and maybe also a verification step (for dev/uat):

```
# Verify the meta tag after edit
- script: |
  grep -q '<meta name="robots" content="noindex, nofollow" />' ./src/index.html || exit 1
  echo "Dev/UAT robots meta tag OK"
  displayName: 'Verify robots meta tag'
```

something like that. What do you think? If you think it could be a valid approach, we should see what other thinks, too

sfarralorenzo (approved):

PR #534, opened 2025-06-24
[PIDM-706-715-pagina-manutenzione]feat: add Maintenance page

forciniti-luigi-jakala:
#### List of Changes

- create MaintenancePage
- create MaintenanceGuard (This guard ensures the route is accessible only when the maintenance feature flag is active)
- create useFeatureFlags hook (replace duplicate flag logic)
- update App.tsx to force redirect to maintenance page if the flag is true
- update Header.tsx  to hide it during maintenance mode

#### Motivation and Context

This change introduces a centralized useFeatureFlags hook to avoid code duplication across the application when evaluating feature flags.
It also ensures that the application properly handles maintenance mode by redirecting users and hiding the header when the corresponding flag is enabled.

Previously, feature flag logic was repeated in both App.tsx and Header.tsx, which made it harder to maintain and extend.
This refactor improves code readability and makes the behavior more consistent.

### Screenshots (Maintenance Mode)
image
### Screenshots (No Maintenance Mode)
image

#### Types of changes

- [X] New feature (non-breaking change which adds functionality)

#### Checklist:

### DIFF
--- a/src/routes/MaintenancePage.tsx
+++ b/src/routes/MaintenancePage.tsx
@@ -37,7 +37,7 @@ export default function MaintenancePage() {
       >
         <img
           src={inProgressIcon}
-          alt="Icona cuore"
+          alt="Work in progress"
           style={{ width: "80px", height: "auto" }}
           aria-hidden="true"
         />


src/App.tsx
```diff
@@ -65,13 +80,37 @@ export function App() {
...
+        featureFlags.enablePspPage,
+        onFeatureFlagError,
+        onFeatureFlagSuccess
+      );
+
+      await evaluateFeatureFlag(
+        featureFlags.enableMaintenance,
+        onFeatureFlagErrorMaintenance,
+        onFeatureFlagSuccessMaintenance
+      );
```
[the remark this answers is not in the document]
forciniti-luigi-jakala: Done

src/App.tsx
```diff
@@ -65,13 +80,37 @@ export function App() {
...
+      "Error while getting feature flag " + SessionItems.enableAuthentication,
+      e
     );
+    setSessionItem(SessionItems.enableMaintenance, false);
+  };
+
+  const onFeatureFlagSuccessMaintenance = (data: { enabled: boolean }) => {
+    setSessionItem(SessionItems.enableMaintenance, data.enabled.toString());
+    setIsMaintenanceEnabled(data.enabled);
+  };
```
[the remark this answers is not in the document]
forciniti-luigi-jakala: Done

src/components/commons/Header.tsx
```diff
@@ -91,36 +90,53 @@ export default function Header() {
...
 
   useEffect(() => {
     void initFeatureFlag();
   }, []);
 
+  // Prevents initial UI render before feature flags are loaded,
+  // avoiding flicker when MaintenancePage should be shown.
+
+  if (loadingFlags) {
+    return <div>Loading</div>;
```
ciuffagianluca, line 137: Maybe is better an empy div rather than a div with "loading" string

sfarralorenzo: another valid option could be the already existing `CheckoutLoader` component

src/components/commons/Header.tsx
```diff
@@ -91,36 +90,53 @@ export default function Header() {
...
-    const storedFeatureFlag = getSessionItem(SessionItems.enableAuthentication);
-    // avoid asking again if you already have received an answer
-    if (!storedFeatureFlag) {
-      await evaluateFeatureFlag(
-        featureFlags.enableAuthentication,
-        onFeatureFlagError,
-        onFeatureFlagSuccess
-      );
+    try {
+      await Promise.all([
```
ciuffagianluca, line 99: Please remove also here the double `checkFeatureFlag` call

forciniti-luigi-jakala: Done! 👍

src/App.tsx
```diff
@@ -49,29 +59,42 @@ export function App() {
...
+  const { checkFeatureFlagAll } = useFeatureFlagsAll();
 
   const initFeatureFlag = async () => {
-    // we need to always evaluate this flag since is stored in the local storage
-    await evaluateFeatureFlag(
-      featureFlags.enablePspPage,
-      onFeatureFlagError,
-      onFeatureFlagSuccess
-    );
+    try {
```
infantesimone, line 71: It would be nice to use fp-ts to refactor this code, but it's not mandatory :-)

forciniti-luigi-jakala: Sure, I'll work on this.

infantesimone:
The PR makes sense to me and clearly the need to revisit how feature flags are managed. However, it’s quite a large PR. For next time, I’d suggest the following approach:

1. Start with a dedicated PR that introduces only the new view component, with its unit tests. This allows reviewers to focus only on the UX aspects of the new component.
2. Follow up with a separate PR that adds the feature flag logic, but without linking it yet to the new component. Both of these PRs would fully retro compatible and could be safely deployed to production.
3. Then create a final PR to connect the component usage to the feature flag, supported by integration tests.
4. I also noticed some unrelated refactoring. That could have been handled in separate PRs as well.

This kind of approach, aligned with continuous delivery principles  it helps maintain high-quality reviews and not just from a technical point of view.

ciuffagianluca:
@infantesimone I agree with your though. In this developing we didn't follow similar approach due to need of going really soon in production and we have a one shot approach to this task.Anyway your workflow would have been applied thinking a little more to the developing strategy. Than you for your sugggestion.

src/App.tsx
```diff
@@ -49,29 +59,42 @@ export function App() {
...
+        const enabled = allFlags.isMaintenancePageEnabled;
+        dispatch(setMaintenanceEnabled({ maintenanceEnabled: enabled }));
+      }
+      if (
+        featureFlags[SessionItems.enableScheduledMaintenanceBannerEnabled] in
+        allFlags
+      ) {
+        const enabled = allFlags.isScheduledMaintenanceBannerEnabled;
+        setSessionItem(
+          SessionItems.enableScheduledMaintenanceBannerEnabled,
```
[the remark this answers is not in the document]
forciniti-luigi-jakala: Since the other feature flags follow the enableX format in the SessionItems enum (e.g., enableAuthentication, enablePspPage), I think it’s more consistent to use enableScheduledMaintenanceBanner instead of isScheduledMaintenanceBannerEnabled.
Let me know if that works for you — happy to align on naming conventions!

sfarralorenzo (dismissed):

pietro-tota (dismissed):
LGMT 🚀  but we should add some unit tests for code coverage

CianoDanilo (approved):

ciuffagianluca:
Bypass some tests failure. They will be addressed in a dedicated PR

PR #562, opened 2025-07-29
fix: [PIDM-608] Login Button Translation

matteo-samoila-jakala:
#### List of Changes

The mui-italia library has been updated and a prop has been added to the login header.

#### Motivation and Context
This PR is made so that the login button in the header gets correctly translated according to the language selected in the footer menu.

Unit tests

image

#### Types of changes

#### Checklist:

### DIFF
--- a/src/components/commons/LoginHeader.tsx
+++ b/src/components/commons/LoginHeader.tsx
@@ -243,6 +243,7 @@ export default function LoginHeader() {
         enableAssistanceButton={false}
         onAssistanceClick={() => {}} // eslint-disable-line @typescript-eslint/no-empty-function
         onLogin={handleClickOnLogin}
+        translationsMap={{logIn: t("authExpiredPage.buttons.login")}}
       />
       <Box display="none">
         <ReCAPTCHA
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "@mui/icons-material": "^5.14.3",
     "@mui/lab": "^5.0.0-alpha.140",
     "@mui/material": "^5.16.14",
-    "@pagopa/mui-italia": "1.5.2",
+    "@pagopa/mui-italia": "1.6.1",
     "@pagopa/openapi-codegen-ts": "^13.2.0",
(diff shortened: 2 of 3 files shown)


src/components/commons/LoginHeader.tsx
```diff
@@ -243,6 +243,7 @@ export default function LoginHeader() {
         enableAssistanceButton={false}
         onAssistanceClick={() => {}} // eslint-disable-line @typescript-eslint/no-empty-function
         onLogin={handleClickOnLogin}
+        translationsMap={{ logIn: t("authExpiredPage.buttons.login") }}
```
sfarralorenzo, line 246: This component is used across multiple contexts since it's embedded directly within the `<Header/>` component. I think that "semantically" speaking it should be `mainPage.header.login` instead of `authExpiredPage.buttons.login`, to be aligned with the existing `mainPage.header.logout` (so note that `mainPage.header.login` doesn't exist yet)

sfarralorenzo (approved):
just a single non-blocking note to think about

src/components/commons/LoginHeader.tsx
matteo-samoila-jakala: great point actually, thanks for pointing it out, it's fixed now

src/components/commons/__tests__/LoginHeader.test.tsx
```diff
@@ -306,4 +308,15 @@ describe("LoginHeader", () => {
...
+  it("renders the login button with translated text", () => {
+    renderWithReduxProvider(
+      <MemoryRouter>
+        <LoginHeader />
+      </MemoryRouter>
+    );
+
+    const loginButton = screen.getByRole("button", { name: "Login" });
+    expect(loginButton).toBeInTheDocument();
+    expect(tMock).toHaveBeenCalledWith("authExpiredPage.buttons.login");
```
infantesimone, line 320: pay attention to renaming

infantesimone (approved):

PR #564, opened 2025-08-05
fix: [PIDM-869] fix tracking mixpanel

FrancescoDiTonnoJakala:
#### Types of changes

#### Checklist:


infantesimone:
PIDM-869 has several issues reported, but I shouldn’t address all of them in the same PR. What do you think about focusing the first PR on fixing the amount and handling the QR error event?

PR #570, opened 2025-09-15
fix: [PIDM-932] center OK/KO pages vertically with flexbox and minHeight

FrancescoDiTonnoJakala:
#### List of Changes
- Removed py: 5 from the wrapper and the content box.
- Added a wrapper with minHeight: '20dvh', display: 'flex', alignItems/justifyContent: 'center'.
- Replaced redundant vertical paddings/margins with gap to manage spacing between elements.

#### Motivation and Context
Currently, the centering relative to the viewport height is fixed, causing problems on small screens: there is an unnecessary space at the beginning that cuts off the main content of the page.

#### Types of changes

#### Checklist:


ciuffagianluca (approved):

src/routes/PaymentResponsePageV2.tsx
```diff
@@ -199,19 +199,21 @@ export default function PaymentResponsePageV2() {
...
         <Box
           sx={{
-            py: 5,
-            display: "flex",
-            justifyContent: "center",
             width: "100%",
+            display: "flex",
             flexDirection: "column",
             alignItems: "center",
+            textAlign: "center",
```
sfarralorenzo, line 215: I don't find this requirement in the task.
The elements should be centered, but here we're also centering the "Puoi dirci com’è andata?" - and all the other - text inside the box, that doesn't seem related to the task

FrancescoDiTonnoJakala: Solved here https://github.com/pagopa/pagopa-checkout-fe/pull/570/commits/fc846672869134a3375b4e83060800c5ef8dd075

ciuffagianluca (approved):

PR #574, opened 2025-09-17
feat: [PIDM-732] added betterstacksdk.js

FrancescoDiTonnoJakala:
#### Types of changes

#### Checklist:

### DIFF
--- a/src/betterstacksdk.js
+++ b/src/betterstacksdk.js
@@ -1,0 +1,170 @@
+if (window.BetterUptimeAnnouncementWidget != null) {
+  window.BetterUptimeAnnouncementWidget.initialize();
+} else {
+  window.BetterUptimeAnnouncementWidget = {
+    currentScript: document.currentScript,
+
+    cookieName: 'betteruptime_announcement_bar_dismissed',
+
+    announcementBarWrapper: null,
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -59,6 +59,7 @@
     <meta http-equiv="Pragma" content="no-cache">
     <script src="env-config.js"></script>
     <script src="npgsdk.js" async></script>
+    <script src="betterstacksdk.js" async></script>
```
sfarralorenzo, line 62:
```suggestion
    <script src="betterstacksdk.js" defer></script>
```
It could be safer to put it at the end of the body to guarantee that is executed after the DOM has been parsed.

If we have to keep it in the `<head>`, It could be safer to use `defer` here. With `async` the script is executed as soon as it is downloaded, possibly blocking the html rendering and parsing.
With `defer` e <head> tag and set them to defer and the loading of the scripts will be deferred until the DOM has been parsed, minimizing the chance to interfere with the flow

src/betterstacksdk.js
sfarralorenzo, line 1: we could consider `src/assets/` adding a `script` folder or similar? Other local scripts in the future will also go there

PR #652, opened 2026-05-26
fix: [PIDM-1957, PIDM-2251] Fix bug back routing to insert mail

RiccardoGiuliani:
#### List of Changes

- [PIDM-1957](https://pagopa.atlassian.net/jira/servicedesk/projects/PIDM/boards/2036?assignee=5ca319aae6f7e8117ffb4b20&selectedIssue=PIDM-1957) Replaced navigate(-1) with explicit navigation in PaymentChoicePage and PaymentNoticePage back button handler.
- [PIDM-2251](https://pagopa.atlassian.net/jira/servicedesk/projects/PIDM/queues/issue/PIDM-2251) Modified regex to accept qr code under 10 cents

#### Motivation and Context

When a user selects a payment method, proceeds to the summary page, and then clicks edit to go back to the method list, the browser history stack gets polluted with intermediate entries (e.g. /inserisci-carta). This caused the back button to navigate to the previously selected payment method page instead of the email form, effectively trapping the user.

Replaced navigate(-1) with an explicit navigation to the root route from PaymentNoticePage: this page is directly accessible via URL, using browser history (navigate(-1)) could navigate the user outside the application (e.g. to an external site or blank tab).

#### Types of changes

#### Checklist:

[PIDM-1957]: https://pagopa.atlassian.net/browse/PIDM-1957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PIDM-2251]: https://pagopa.atlassian.net/browse/PIDM-2251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pietro-tota (changes requested):
on hold: this back button behavior was intentionally to handle pages navigation, why change this?

pietro-tota (dismissed):

pietro-tota (dismissed):

forciniti-luigi-jakala (dismissed):

src/routes/PaymentSummaryPage.tsx
```diff
@@ -143,7 +143,9 @@ export default function PaymentSummaryPage() {
         idCancel="paymentSummaryButtonBack"
         disabledSubmit={false}
         handleSubmit={onSubmit}
-        handleCancel={() => navigate(-1)}
+        handleCancel={() =>
+          navigate(`/${CheckoutRoutes.INSERISCI_DATI_AVVISO}`)
+        }
```
dylantangredi-jakala, line 150: I tested the qr code scanning in dev with this branch and the back button on `/dati-pagamento` goes to `/inserisci-dati-avviso`, it does not go to back to the qr scan page `/leggi-codice-qr` like it still does in UAT/PROD. Is this behavior expected or a regression in some way?

src/routes/PaymentPspListPage.tsx
```diff
@@ -250,7 +250,7 @@ export default function PaymentPspListPage() {
             disabledSubmit={!submitEnabled}
             handleSubmit={onSubmit}
             handleCancel={() => {
-              navigate(-1);
+              navigate(`/${CheckoutRoutes.SCEGLI_METODO}`);
             }}
```
dylantangredi-jakala, line 254: `/lista-psp` is entered both from `/scegli-metodo` (`PaymentChoice.tsx:121,146`) and from `/inserisci-carta` (`IframeCardForm.tsx:104`, when `enablePspPage` is on). 

With this change we hardcode `scegli-metodo` and this sends card users back past the card form, losing the data they entered. I tested it in DEV, while in UAT and PROD when we go back we go to the card data form as expected, so it can be considered a regression.

Could this reuse the same `enablePspPage` check, or receive the origin via navigate state maybe?

dylantangredi-jakala (commented):
found another regression, we should check every "cancel/back" button behavior after the change. If these canged behavior are expected and documented, we can ignore my comments and proceed with the merge

src/routes/PaymentSummaryPage.tsx
RiccardoGiuliani: To properly navigate back to `/leggi-codice-qr` for QR users while still going to `/inserisci-dati-avviso` for manual entry users, we would need to either:
1. Pass the origin via navigate state from `PaymentNoticePage` — but this state is lost on page refresh and would need to be propagated through `PaymentEmailPage` as well (since the user can reach `/dati-pagamento` back from `/inserisci-email` too)
2. Store the origin in `sessionStorage` — adding more session state to maintain and clean up

Both approaches add complexity for a relatively minor UX difference (the user lands on `/inserisci-dati-avviso` with all input compilated). Do you think we need to implement this fix?

@pietro-tota cc

pietro-tota: IMHO we could perform navigation to /inserisci-dati-avviso in the back navigation avoiding add complexity

@infantesimone @CianoDanilo @ciuffagianluca what do you think?

ciuffagianluca: I agree with @pietro-tota

ciuffagianluca (dismissed):

src/routes/PaymentPspListPage.tsx
dylantangredi-jakala: uhm with the new commit [a6e7cef](https://github.com/pagopa/pagopa-checkout-fe/pull/652/commits/a6e7cef80ed3b0017296f76f9be161db09dcf5f7) I see that the new condition is  `paymentMethod?.paymentTypeCode === "CP"`, which Im not sure is equivalent to "arrived from `/inserisci-carta`". 

Do the users who use saved wallet also carry `paymentTypeCode === "CP"`? Because in this case a saved card user pressing the back button is now sent to `/inserisci-carta`, a blank new card form they never visited

src/utils/regex/validators.tsx
```diff
@@ -17,7 +17,7 @@ export function digitValidation(text: string) {
 }
 
 export function qrCodeValidation(code: string) {
-  return /^[a-zA-Z]{6}(\|[0-9]{3})(\|[0-9]{18})(\|[0-9]{11})(\|[0-9]{2,11})$/.test(
+  return /^[a-zA-Z]{6}(\|[0-9]{3})(\|[0-9]{18})(\|[0-9]{11})(\|[1-9][0-9]{0,10})$/.test(
```
dylantangredi-jakala: we now accept eurocents amounts under 10 but not the zero padded values, is this expected? e.g. "05" would be not accepted

src/routes/PaymentPspListPage.tsx
dylantangredi-jakala: Ok we can reverto to `SCEGLI_METODO` route as @pietro-tota agrees that is the most appropriate way to handle this. I was just worried about changing the behavior from what it is established in prod, sorry

src/utils/regex/validators.tsx
dylantangredi-jakala: ok we can leave the zero padding unaccepted so only 1-9 are accepted (for single digits), resolving this

pietro-tota (approved):

dylantangredi-jakala (approved):

PR #654, opened 2026-06-10
feat: [CHK-4860] handle retry after header

ivanferrando-pagopa:
#### List of Changes

- Refactored outcome polling logic to handle Retry-After headers
- Added new unit test for retry-after case
- Updated polling section on readme

#### Motivation and Context

The goal is to handle a possible a `429 Too Many Requests` response with a `Retry-After` header from the `GET /outcomes` API, sending the next request once the required time has elapsed

- Unit tests
- Local & DEV tests

429 response:
image

Following request after the time specified in the Retry-After has elapsed:
image

#### Types of changes

- [X] New feature (non-breaking change which adds functionality)

#### Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

### DIFF
--- a/src/utils/api/response.ts
+++ b/src/utils/api/response.ts
@@ -8,6 +8,7 @@ import * as T from "fp-ts/Task";
 import { pipe } from "fp-ts/function";
 import { DeferredPromise } from "@pagopa/ts-commons/lib/promises";
 import { Millisecond } from "@pagopa/ts-commons/lib/units";
+import { createCounter } from "../../utils/counter";
 import { ecommerceTransactionOutcome } from "../transactions/transactionHelper";
 import {
   exponentialPollingWithPromisePredicateFetch,
@@ -41,6 +42,8 @@ export const decodeToUUID = (base64: string) => {
   return hexToUuid(bytes.toString("hex")).replace(/-/g, "");
 };
 
+const counter = createCounter();
+
 const parseRetryAfterToMs = (
   headerValue: string | null,
   fallbackMs: number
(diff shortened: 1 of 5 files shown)


src/utils/api/response.ts
```diff
@@ -39,21 +41,43 @@ export const decodeToUUID = (base64: string) => {
...
 const ecommerceClientWithPollingV1: EcommerceClientV1 = createClientV1({
   baseUrl: config.CHECKOUT_PAGOPA_APIM_HOST,
-  fetchApi: exponetialPollingWithPromisePredicateFetch(
+  fetchApi: exponentialPollingWithPromisePredicateFetch(
     DeferredPromise<boolean>().e1,
     retries,
     delay,
     timeout,
-    async (r: Response): Promise<boolean> => {
-      counter.increment();
```
CianoDanilo, line 52: Why are we removing the counter?

ivanferrando-pagopa: The number of attempts should already be controlled by the `withRetries` function on fetch.ts

ciuffagianluca: This parameter limits the number of the total attempts, to stop it after configured limit.

ivanferrando-pagopa: Re-added since it's needed to limit the attempts in case of a page refresh

sterondinella (approved):

ciuffagianluca (approved):

CianoDanilo (approved):

PR #659, opened 2026-06-23
feat: [PIDM-2152] add npg sdk sync pipelines

dylantangredi-jakala:
depends on https://github.com/pagopa/pagopa-checkout-fe/pull/658
depends on https://github.com/pagopa/pagopa-infra/pull/3937
depends on https://github.com/pagopa/pagopa-ecommerce-payment-methods-handler/pull/42

#### List of Changes

- add `pagopa-npg-sdk-sync-deploy-pipelines.yml`: hourly scheduled pipeline to download the NPG SDK from Nexi, compute its SHA-384 integrity hash, and upload both to `$web/npg/sdk/`
- add `azure-templates/sync-npg-sdk-steps.yml`: reusable `AzureCLI@2` steps for download, HTTP/size validation, hash computation, blob upload, and CDN purge
- handle the `cdn`/`afd` az-core command deprecation via `extension.use_dynamic_install` before the purge
- add `--exclude-path npg/sdk` to the DEV/UAT/PROD `az storage blob sync` commands so deploys don't delete sync-managed files
- repoint DEV `checkout_npg_sdk_url` to `https://dev.checkout.pagopa.it/npg/sdk/hfsdk.js` and add `checkout_npg_sdk_integrity_url`
- add verify-before-publish origin validation: fetch NPG's source-of-truth hash from the payment-methods-handler integrity endpoint (via APIM) and publish only when it matches the computed `sha384` hash, else fail and keep the last-known-good served
- emit a best-effort `NpgSdkSyncSuccess` customEvent heartbeat to App Insights at the end of a fully successful run (download → verify → publish → purge) so a staleness alert (https://github.com/pagopa/pagopa-infra/pull/3929) can detect the sync stopping or failing; connection string resolved at runtime via `az monitor app-insights component show` (two new template params `app_insights_name`/`app_insights_rg`)
- harden the heartbeat: use `itemsAccepted` from the ingestion response as the authoritative success signal (not curl exit), bound `az`/`curl` with timeouts so telemetry can never hang the job, and keep the instrumentation key out of logs/argv (`set +x`, temp-file payload)

#### Motivation and Context

Part of self-hosting the NPG SDK with Subresource Integrity. Adds the scheduled job that keeps the self-hosted SDK and its integrity hash in sync on our CDN storage, and repoints DEV to load the SDK from our own origin. The `--exclude-path npg/sdk` on the deploy sync commands prevents deploys from wiping the sync-managed files. 

The pipeline validates the SDK it downloads from Nexi against NPG's source-of-truth hash before publishing (fetched from the payment-methods-handler integrity endpoint exposed in https://github.com/pagopa/pagopa-infra/pull/3937), so an unexpected or tampered SDK is never published; the pipeline holds no NPG credential. On a fully successful run it emits a `NpgSdkSyncSuccess` heartbeat to App Insights, which a dead-man's-switch staleness alert (pagopa-infra #3929) uses to detect the hourly sync stopping or failing.

- manual run of `pagopa-checkout-fe.npg-sdk-sync.deploy` from this branch synced `hfsdk.js` + `hfsdk.integrity.json` to DEV `$web/npg/sdk/`; both return 200 and the served SDK matches the published `sha384` hash
- verify-before-publish confirmed in DEV: the run downloaded the SDK (36515 bytes), fetched the source-of-truth hash from the handler integrity endpoint, matched the computed `sha384`, and published; a direct call to the endpoint returns 200 with `integrityHash`
- App Insights heartbeat validated end-to-end in DEV: a successful sync emitted `NpgSdkSyncSuccess` (HTTP 200, `itemsAccepted=1`); the pagopa-infra #3929 staleness alert fired when heartbeats aged out (>3h) and auto-resolved once a fresh heartbeat landed
- DEV deployment from branch `PIDM-2152-add-npg-sdk-sync-pipelines`:
  - `--exclude-path npg/sdk` verified: after deploy, `curl -I https://dev.checkout.pagopa.it/npg/sdk/hfsdk.js` returns 200, integrity JSON still accessible
  - existing assets unaffected: `/npg/style.css` and `/npg/card-default.svg` still return 200 after deploy
  - full card payment flow tested successfully with SDK loading from self-hosted URL with SRI
- no regressions observed in other checkout flows
- https://dev.azure.com/pagopaspa/pagoPA-projects/_build/results?buildId=279742&view=results

#### Types of changes

#### Checklist:


.devops/azure-templates/sync-npg-sdk-steps.yml
```diff
@@ -0,0 +1,76 @@
...
+        if [ "${HTTP_CODE}" != "200" ]; then
+          echo "##vso[task.logissue type=error]SDK download failed with HTTP ${HTTP_CODE}"
+          exit 1
+        fi
+
+        FILE_SIZE=$(stat -c%s /tmp/hfsdk.js)
+        if [ "${FILE_SIZE}" -lt 1024 ]; then
+          echo "##vso[task.logissue type=error]SDK file suspiciously small (${FILE_SIZE} bytes)"
+          exit 1
+        fi
```
pietro-tota:
```suggestion
```

redundant check
+1 for the file size logging but I would remove an hardcoded file size check since we are checking for sha matching later

dylantangredi-jakala: ok so good point on the logging. But warning the "checking sha later" part, we're not actually comparing against a known-good hash anywhere. The pipeline generates the hash from whatever it downloads, and the frontend just enforces that same self-generated hash. So SRI only protects the CDN→browser hop (tamper detection after publish), not the download itself.

That makes the size check not redundant as it's the only guard against Nexi returning a truncated/bad 200, which would otherwise get mirrored and published with a perfectly valid SRI hash.

We can drop hardcoded 1024 (it's an arbitrary number) but I'd swap it for a non-empty check  and keep the size log. What do you think?

Separately, we could discuss on whether we want real origin validation (pin an expected hash + fail-closed but keep last-good) that I am missing in this whole implementation

pietro-tota: ``` we're not actually comparing against a known-good hash anywhere``` -> this is not true, or at least there is a problem in this pr: you should get sha hash from NPG servers with b2b api call and then compare with the calculated one

dylantangredi-jakala: yes thats what I meant, working on it

dylantangredi-jakala: addressed by calling the handler integrity api through apim ->[5405f08](https://github.com/pagopa/pagopa-checkout-fe/pull/659/commits/5405f082d92f6e4371bad9cc52c3000cf0451dc2)

dylantangredi-jakala:
Superseded by https://github.com/pagopa/pagopa-checkout-fe/pull/661 - re-cut supply-only on main (sync pipeline only, no SDK-URL repoint)

PR #663, opened 2026-07-24
[PIDM-2241] fix bin value in calculateFees

FrancescoDiTonnoJakala:
#### List of Changes
- The `retrieveCardData` logic was moved to before the redirect to the PSP selection page and made asynchronous in `IFrameCardForm`

#### Motivation and Context
this resolves the bug where the BIN was not populated in the card method in post Fees

- Unit Test

#### Types of changes

#### Checklist:


src/features/payment/components/IframeCardForm/IframeCardForm.tsx
```diff
@@ -118,6 +115,11 @@ export const retrievePaymentSessionFn = (deps: RetrievePaymentSessionDeps) => {
       );
     },
   });
+
+  if (localStorage.getItem(SessionItems.enablePspPage) === "true") {
```
pietro-tota: assign this localStorage.getItem(SessionItems.enablePspPage) === "true" to a boolean value once and use in both if codes, to make code clean and readable

pietro-tota (dismissed):
LGMT, just some cosmetic consideration.

Please add some test to add coverage

src/features/payment/components/IframeCardForm/IframeCardForm.tsx
FrancescoDiTonnoJakala: Solved here https://github.com/pagopa/pagopa-checkout-fe/pull/663/changes/1f4ec40cf0e2b20a91414e1fce386c2c68db8ec3

svariant (dismissed):

src/features/payment/components/IframeCardForm/IframeCardForm.tsx
```diff
@@ -97,19 +99,17 @@ export const retrievePaymentSessionFn = (deps: RetrievePaymentSessionDeps) => {
...
-  if (localStorage.getItem(SessionItems.enablePspPage) === "true") {
-    setLoading(false);
-    navigate(`/${CheckoutRoutes.LISTA_PSP}`);
-    return;
-  }
+  const isPspPageEnabled =
+    localStorage.getItem(SessionItems.enablePspPage) === "true";
 
-  void retrieveCardData({
+  await retrieveCardData({
```
dylantangredi-jakala, line 105: The approach is right and I confirmed on DEV that the happy path now sends bin in `POST /fees`. The issue is that the navigation is not gated on `retrieveCardData` succeeding.

`retrieveCardData` never rejects: every failure path calls onError, returns `{}`, and still resolves (`ecommercePaymentMethodsHelper.ts:111-149`). So await tells us the call completed, not that it worked, and the trailing `if (isPspPageEnabled)` runs unconditionally.

Reproduced on DEV with `isPspPickerPageEnabled forced` on and `GET /ecommerce/checkout/v1/payment-methods/*/sessions/*` blocked (see screenshot):
- `onError` fires (CONNECTION then SERVER), but no error modal is ever shown
- we land on `/lista-psp` regardless, with no feedback to the user
- `POST /v2/payment-methods/{id}/fees` goes out with no bin

We could gate the navigation on a decoded bin inside `onResponseSessionPaymentMethod`, mirroring the check already in `handleSessionPaymentMethodResponse`. That also removes the need for async/await, since `retrieveCardData` invokes the callback synchronously just before resolving. What do you think?

something like for example 

```
onResponseSessionPaymentMethod: (resp) => {
   if (isPspPageEnabled) {
     return pipe(
       SessionPaymentMethodResponse.decode(resp),
       O.fromEither,
       O.chain((r) => O.fromNullable(r.bin)),
       O.fold(
         () => onError(ErrorsType.GENERIC_ERROR),
         () => {
           setLoading(false);
           navigate(`/${CheckoutRoutes.LISTA_PSP}`);
         }
       )
     );
   }
   handleSessionPaymentMethodResponse(resp, onError, onSuccess, onPspNotFound);
 },
```

Image

src/features/payment/components/IframeCardForm/__tests__/IframeCardForm.test.tsx
```diff
@@ -958,12 +958,24 @@ describe("retrievePaymentSessionFn", () => {
     localStorageMock.clear();
   });
 
-  it("should navigate to LISTA_PSP and skip retrieveCardData when enablePspPage is true", () => {
+  it("should await retrieveCardData, skip getFees, and navigate to LISTA_PSP when enablePspPage is true", async () => {
```
dylantangredi-jakala, line 961: This covers only the happy path, but the behaviour this PR changes is what happens when `retrieveCardData` fails (see main comment on the issue). 

Could we add two cases with `enablePspPage=true`: one where the mock calls `onError` and one where the response has no bin, asserting we do not navigate to `/lista-psp` in either?

src/features/payment/components/IframeCardForm/IframeCardForm.tsx
FrancescoDiTonnoJakala: Solved here https://github.com/pagopa/pagopa-checkout-fe/pull/663/changes/0df9535535d7c9ddd0574824926dcd11cc150413

src/features/payment/components/IframeCardForm/__tests__/IframeCardForm.test.tsx
FrancescoDiTonnoJakala: Solved here https://github.com/pagopa/pagopa-checkout-fe/pull/663/changes/53e684ad2e46d6a87642d3c8dfec24a63b103b08

dylantangredi-jakala (approved):
Good work!

pietro-tota (approved):

